PR #32, opened 2016-01-27
Support es6 template strings (`` delimeters)

mogelbrod:
This PR adds [es6 template string](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/template_strings) support to the parser.

I also added some code that removes dot-separators from expressions to prevent them from nesting.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,7 +100,7 @@ Parser.prototype._transform = function(file, encoding, done) {
     // and we parse for functions...
     // =============================
     var fnPattern = '(?:' + this.functions.join( ')|(?:' ).replace( '.', '\\.' ) + ')';
-    var pattern = '[^a-zA-Z0-9_](?:'+fnPattern+')(?:\\(|\\s)\\s*(?:(?:\'((?:(?:\\\\\')?[^\']*)+[^\\\\])\')|(?:"((?:(?:\\\\")?[^"]*)+[^\\\\])"))';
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -174,6 +174,12 @@ Parser.prototype._flush = function(done) {
     // ==========================
     for (var index in self.translations) {
         var key = self.translations[index];
+
+        // simplify ${dot.separated.variables} into just their tails (${variables})
+        key = key.replace(/\$\{([^}]+)\}/g, function(m, expr) {
```
karellm: What is the purpose of simplifying these? Also this could be written with a single replace and regex.

karellm:
Thanks for the PR, I added a question in the code. Could you please add tests and update the README?

index.js
mogelbrod: The primary reason to remove any dotted paths is to avoid breaking them up when constructing the json object, for example:
```
t(`root.${dot.separated}`)
// JSON file after parsing:
{"root": {"${dot": {"separated}": ""}}
```

This could of course be solved by modifying how the object is constructed, but I found that most times dot-separated expressions were present inside lookups they looked something like this:
```
t(`root.${this.state.account.type}`)
```
With `type` (and arguably `account`) being the only parts relevant to the translation.

mogelbrod:
Thanks for the quick response!
Good catch on the regex, fixed it now.
Also added a test and readme note.
Anything else missing?

karellm:
@mogelbrod Thanks for the contribution, I just published `0.5.0` with your changes!

mogelbrod:
Great, really appreciate it!

PR #33, opened 2016-02-01
Add keepRemoved option (cli and gulp) + expose writeOld option to cli

mogelbrod:
* Add `keepRemoved` (`--keep-removed`) option which, when enabled, prevents the parser from removing translation keys present in the existing json files when not found in the source code.
* Expose `writeOld` option to cli as `--no-old` (since default is to write them).
* Add `keepRemoved` option test (part of `mergeHash`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Thanks a lot to all the previous [contributors](https://github.com/i18next/i18ne
 - **-l, --locales <list>**: The locales in your applications. Defaults to `en,fr`
 - **--directoryFilter**: Globs of folders to filter
 - **--fileFilter**: Globs of files to filter
+- **--keep-removed**: Prevent keys no longer found from being removed
+- **--write-old false**: Avoid saving the \_old files
 
 ---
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -64,6 +64,8 @@ Thanks a lot to all the previous [contributors](https://github.com/i18next/i18ne
 - **-l, --locales <list>**: The locales in your applications. Defaults to `en,fr`
 - **--directoryFilter**: Globs of folders to filter
 - **--fileFilter**: Globs of files to filter
+- **--keep-removed**: Prevent keys no longer found from being removed
+- **--no-old**: Avoid saving the _old files
```
karellm: I prefer we don't multiply the APIs. There is already a `writeOld` option doing the same thing.

For backward compatibility and consistency, can you make that `writeOld`?

karellm:
Thanks for the PR, I have added one comment. 

After it is fixed, can you squash your commits into one? I will then merge and deploy a new version.

README.md
mogelbrod: I completely agree. Since _old file writing is the default, am I correct in assuming it would be disabled with `i18next --write-old=false`?
If this is your preference I'll fix it right away.

mogelbrod:
Thanks for the quick response as usual :)
Absolutely!

README.md
karellm: Exactly! I prefer your name but I don't think this justifies to break backward compatibility.

mogelbrod:
Everything should be set and ready for merging now.

README.md
karellm: I think the README needs to be updated. `no-old` should be `writeOld`. Maybe even: `**--write-old false**: Avoid saving the _old files`.

mogelbrod:
Fixed the remaining README error now.

karellm:
@mogelbrod Thanks for the contribution! Just published `0.6.0` :)

PR #34, opened 2016-02-13
Add custom html attributes option to cli + gulp

sifrenette:
Added an additional option in gulp and in the cli (`--attributes`) to be able to parse custom html attributes.


karellm:
Wow, a documented, tested and squashed PR :smiley:  Thanks for the contribution!

PR #40, opened 2016-05-12
Display an error when the matches function's first arg is not a string literal

laszlopandy:
If someone tried to call the translation function with a variable, currently the parser will fail silently. This makes it an error to call the translation function on a string which cannot be extracted (and thus will never be translated).


karellm:
Thanks for the contribution, I published `0.8.0` with the changes.

laszlopandy:
Thanks for the quick release!

PR #42, opened 2016-06-17
Handle strings with newlines, tabs and backslashes in them.

laszlopandy:
Previously we sending translators `"Strings with:\\ndouble-escaped newlines"`. This fixes the issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,8 +153,13 @@ Parser.prototype._transform = function(file, encoding, done) {
     // finally we add the parsed keys to the catalog
     // =============================================
     for (var j in keys) {
+        var key = keys[j];
         // remove the backslash from escaped quotes
-        var key = keys[j].replace(/\\('|"|`)/g, '$1');
+        key = key.replace(/\\('|"|`)/g, '$1')
+        key = key.replace(/\\n/g, '\n');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -153,8 +153,13 @@ Parser.prototype._transform = function(file, encoding, done) {
     // finally we add the parsed keys to the catalog
     // =============================================
     for (var j in keys) {
+        var key = keys[j];
         // remove the backslash from escaped quotes
-        var key = keys[j].replace(/\\('|"|`)/g, '$1');
+        key = key.replace(/\\('|"|`)/g, '$1')
+        key = key.replace(/\\n/g, '\n');
```
karellm: Is there a reason your are not simply doing `key = key.replace(/\\/g, '\');`?

karellm:
Thanks for the PR! I added a comment but other it looks good.

index.js
laszlopandy: `\n` is a single char, so it `\r` and `\t`. You cannot match on the `\` part because it only exists in the pre-parsed source.

karellm:
@laszlopandy I just published `0.9.0`. Thanks for the contribution!

PR #44, opened 2016-06-27
Generate new translationHash object for each locale.

Mankee:
The for loop that generates the translationHash object needs to run for each locale, otherwise you get translation values from whatever locale had a matching key value pair for that translation.


karellm:
@Mankee Could you write a test to show what was failing before?

karellm:
Closing for lack of activity, feel free to reopen with proper testing

PR #47, opened 2016-07-13
Add option to ignore variable check in i18n.t and continue

birdy-:


birdy-:
Somebody here ? :)

karellm:
`0.10.0` was just released

o5k:
This only offers `ignoreVariables` for `CLI` usage, but there is still no way to `ignoreVariables` with `gulp`..

karellm:
@o5k I added it to the README

PR #48, opened 2016-07-13
Add parsing of context in option of translation function

birdy-:


birdy-:
Oops tests fails.

birdy-:
Better now ;)

Issue #46, opened 2016-07-08
Doesn't handle context via { context: 'male' }


birdy-:
We are trying to keep our translations in "real" english to have a acceptable fallback if a string isn't translated (`Do you want to continue ?` looks better than `modal.question.continue` when the string is not available in your language at the moment).
i18next doesn't handle `i18n.t('From', { context: 'date' })` but `i18n.t('From::date')`. It's an issue when you rely on key for fallback.
I quickly look at the code, and it seems we don't get any options atm.

jamuhl:
should work...like http://i18next.com/translate/keyBasedFallback/

birdy-:
It seems context is not stripped from the "key" if the key includes context directly :
```
i18next.t('My key', { context: 'mycontext' }); // output 'My key'
i18next.t('My key::mycontext'); // output 'My key::mycontext'
```

jamuhl:
did you change the context seperator? Else it should be `i18next.t('My key_mycontext')`

karellm:
First of all, I'm maintaining this project because there are a lot of users but I'm not building new features. Basically if you write a PR, I will merge it but can't put development time at the moment.

@birdy- The way the parser work, it should prepopulate other language with the default translation of your first locale (see https://github.com/i18next/i18next-parser/issues/43). This might address your concern. I personally think you should never use actual english because if you change the text for whatever reason (typo and a change of mind), it will be painful to consolidate.

As to matching `keyBasedFallback`, I agree that the parser should match `i18n` options as closely as possible. 

Finally for the context issue, as @jamuhl puts it, the default separator is `_` and not `::`. There is not option to change it but could be added pretty easily if you want to tackle it.

birdy-:
The problem isn't the ctxSeparator (it could be `:` or `_` or `d-.-b` for some fun). It's just a missing feature which could be very useful in my case.
I totally understand it would be some work to handle this since we don't parse options atm. I posted this issue mainly to see if I get some :+1:. I'll try to push a PR this week.

birdy-:
Done in the PR above.

PR #49, opened 2016-07-13
Add parsing of context in option of translation function

birdy-:
- Regex has been updated to match options
- If a `context` is found, it will define key as `key_context`
- add param to change context separator (`_` by default)

Solves https://github.com/i18next/i18next-parser/issues/46

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Thanks a lot to all the previous [contributors](https://github.com/i18next/i18ne
 - **-n, --namespace <string>**: Default namespace used in your i18next config. Defaults to `translation`
 - **-s, --namespace-separator <string>**: Namespace separator used in your translation keys. Defaults to `:`
 - **-k, --key-separator <string>**: Key separator used in your translation keys. Defaults to `.`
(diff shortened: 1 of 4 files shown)


birdy-:
Somebody here ? :)

karellm:
Been busy but I will review this later today 👍

birdy-:
Yes ! And see the other PR ;)

karellm:
@birdy- You are making me review the only part I always pray people don't make a PR about... regexes 😢 

It should be released soon, I'm reviewing it now.

karellm:
@birdy- I really like how you used the ungready operator 👍 

My only concern is that the way you parse the object, the `context` has to be the last key. Can the object take other attributes?

karellm:
`0.10.0` was just released

birdy-:
Yes, I should fix that. I think it can be solved easily

birdy-:
And thanks for merge(s) btw :)

birdy-:
Oh you already fixed it ;) Ty :)

PR #51, opened 2016-08-07
Hotfix/issue50

arjanvanderleden:
In this PR 
- a fix for issue 50 : in jade file attributes preceded with ( are not matched
- a fix for single quotes attributes not being parsed, added a new option 'includeSingleQuoted' that enables this fix
- changes in test/parser.js in order to run test on the windows platform
- added 2 tests in parser.js to test both fixes

### DIFF
--- a/test/parser.js
+++ b/test/parser.js
@@ -11,12 +11,12 @@ var replaceEmpty    = helpers.replaceEmpty
 
 describe('parser', function () {
 
-    function isFilenameMatch(firstFile,secondFile){
+    function isFilenameMatch(firstFile,secondFile) {
         return (firstFile === secondFile) || firstFile.replace('\\','/') === secondFile.replace('\\','/');
     }
 
-    function isFilenameInArray(filename,filenameArray){
-        return (filenameArray.indexOf(filename)>-1) || filenameArray.map(function(fn){return fn.replace('\\','/')}).indexOf(filename.replace('\\','/'))>-1;
+    function isFilenameInArray(filename,filenameArray) {
+        return (filenameArray.indexOf(filename)>-1) || filenameArray.map(function(fn) {return fn.replace('\\','/')}).indexOf(filename.replace('\\','/'))>-1;
     }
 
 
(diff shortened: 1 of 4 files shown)


karellm:
@arjanvanderleden I will not review the PR until you preserve the original styleguide of the project. 

I agree that the code is not perfectly consistent because it is the result of many contributions at this point. Anything that brings consistency while preserving the existing style is welcome. Anything that redefines the styleguide is not. 

For example, removing aligment of `=` on variable definitions is not something I want to merge. Fixing spaces in `if (` is fine.

arjanvanderleden:
@karellm I'm sorry about that. If you could point me to the style guide, that would be a big help.

karellm:
@arjanvanderleden no worries, I don't have a strict guideline for this project because it was written before the [styleguide I stand by](https://github.com/pillars/styleguides/blob/master/javascript.md). The best is that you try to change as little things as possible. I'm fine that we switch to no spaces inside parentheses (I saw you changes these) though please don't change the indentation. Also I would much prefer a separate PR for style fixes so it doesn't prevent merging the bugfix you wrote.

And I forgot the most important, thanks a lot for contribution!

arjanvanderleden:
I see that code formatting of my editor removed the alignment of the options. I will revert and create a separate PR for the style changes

test/parser.js
```diff
@@ -1,4 +1,25 @@
+var fs              = require('fs');
```
karellm: Why did you add this?

test/parser.js
```diff
@@ -7,7 +28,8 @@ describe('parser', function () {
         });
 
         i18nextParser.on('data', function (file) {
-            if ( file.relative === 'en/translation.json' ) {
+            if (isFilenameMatch(file.relative,'en/translation.json')) {
+            //if ( file.relative === 'en/translation.json' ) {
```
karellm: Please do not let commented out code in your commits

test/parser.js
arjanvanderleden: to debug test/parser.js the variables declared in test/test.js were included in parser.js
(debugging is no possible via test.js, due to `eval(fs.readFileSync(__dirname+'/parser.js')+'');`)

arjanvanderleden:
superseded by #52

PR #52, opened 2016-08-25
Fix pattern: ignore functions which terminate with a 't'

birdy-:
Atm, we look for something like:
(anything)t(spaces or nothing)(a `(` or nothing)...
Sadly, if we have `import './lol.scss';` it will be detected as a translation (`t 'lol.scss'`)

For fix, I added a check on the previous char:
(start of the line or anything except [a-zA-Z0-9_])t(spaces or nothing)(a `(` or nothing)...

Now it works.


karellm:
Thanks!

karellm:
`0.10.1` was just published

PR #56, opened 2016-10-28
Add --track-paths

sourceRS:
This PR adds the ability to call the parser with `--track-paths true`, resulting in an additional file being output to `prefix + namespace + suffix + '.po' + extension` (ex: locales/en/translation**.po**.json in addition to locales/en/translation.json) with a slightly different format than the original, like so:

```json
// .json
{
    "keyOne": "a value",
}

// .po.json
{
    "keyOne": {
        "msgstr": "a value",
        "paths": ["where/a/value/was/found.html"]
    }
}
```

or with plurals and nested keys:

```json
// .json
{
    "keyOne": {
        "keyTwo": "I have to key",
        "keyTwo_plural": "I have {count} keys",
    }
}

// .po.json
{
    "keyOne": {
        "keyTwo": {
            "msgstr": "I have to key",
            "paths": ["fake/path/index.jade"]
        },
        "keyTwo_plural": "I have {count} keys",
    }
}

```

Additionally, there is a [PR](<PR URL>) up for the gettext-converter to handle both formats when converting from .json -> .po

### DIFF
--- a/index.js
+++ b/index.js
@@ -250,7 +250,7 @@ Parser.prototype._flush = function(done) {
                 }
                 catch (error) {
                     this.emit( 'json_error', error.name, error.message );
-                    currentTranslations = {};
+                    oldTranslations = {};
                 }
             }
             else {
--- a/index.js
+++ b/index.js
@@ -37,6 +37,7 @@ function Parser(options, transformConfig) {
     this.writeOld           = options.writeOld !== false;
     this.keepRemoved        = options.keepRemoved;
     this.ignoreVariables    = options.ignoreVariables || false;
+    this.trackPaths         = options.trackPaths || false;
 
     ['functions', 'locales'].forEach(function( attr ) {
         if ( (typeof self[ attr ] !== 'object') || ! self[ attr ].length ) {
@@ -88,6 +89,15 @@ Parser.prototype._transform = function(file, encoding, done) {
     // =================
     if(file.isBuffer()) {
         data = file.contents;
+    }
+
+
+
+    // we find the relative path
+    // =========================
+    var relativeFilePath = path.relative(__dirname, this.base);
+    if (file.path) {
+        relativeFilePath = path.relative(process.cwd(), file.path);
     }
 
 
@@ -176,7 +186,10 @@ Parser.prototype._transform = function(file, encoding, done) {
             key = key.replace( self.namespaceSeparator, self.keySeparator );
         }
 
-        self.translations.push( key );
+        self.translations.push({
+            'key': key,
+            'paths': [relativeFilePath]
+        });
     }
 
     done();
@@ -189,21 +202,57 @@ Parser.prototype._flush = function(done) {
     var self = this;
     var base = path.resolve( self.base, self.output );
     var translationsHash = {};
+    var translationsHashCommented = {};
 
 
+
+    // merge file paths
+    // ================
+    var tempHash = {};
+    _.map(self.translations, (_, index) => {
+        var translation = self.translations[index];
+
+        if (tempHash.hasOwnProperty(translation.key)) {
+            tempHash[translation.key] = tempHash[translation.key].concat(translation.paths);
+        } else {
+            tempHash[translation.key] = translation.paths;
+        }
+
+        // keep file paths in predictable order
+        tempHash[translation.key].sort();
+
+        return _;
+    });
 
     // remove duplicate keys
     // =====================
-    self.translations = _.uniq( self.translations ).sort();
+    self.translations = _.map(tempHash, (paths, key) => {
+        return { key: key, paths: paths};
+    });
+
+    // sort by key
+    // ===========
+    self.translations = _.sortBy(self.translations, 'key' );
 
 
 
-    // turn the array of keys
+    // turn the array of objects
     // into an associative object
     // ==========================
+    // from
+    // [{ keyOne: '', paths: ['', '']}, { keyTwo: '', paths: ['']}]
+    // to
+    // { 'keyOne': { msgstr: '', paths: ['', ''] }, keyTwo: { msgstr: '', paths: [''] } }
+    // ==================================================================================
     for (var index in self.translations) {
         // simplify ${dot.separated.variables} into just their tails (${variables})
-        var key = self.translations[index].replace( /\$\{(?:[^.}]+\.)*([^}]+)\}/g, '\${$1}' );
+        var key = self.translations[index].key.replace( /\$\{(?:[^.}]+\.)*([^}]+)\}/g, '\${$1}' ),
+            paths = self.translations[index].paths;
+
+        if (self.trackPaths) {
+            translationsHashCommented = helpers.hashFromString( key, self.keySeparator, translationsHashCommented, paths );
+        }
+
         translationsHash = helpers.hashFromString( key, self.keySeparator, translationsHash );
     }
 
@@ -289,7 +338,26 @@ Parser.prototype._flush = function(done) {
                 self.push( mergedOldTranslationsFile );
             }
 
+            if ( self.trackPaths ) {
+                var namespacePathCommented = path.resolve(
+                    localeBase,
+                    prefix + namespace + suffix + '.po' + extension
+                );
 
+                // merges existing translations with the new commented ones
+                mergedTranslationsCommented = helpers.mergeHash( currentTranslations, translationsHashCommented[namespace], null, this.keepRemoved );
+
+                // restore old translations if the key is empty
+                mergedTranslationsCommented.new = helpers.replaceEmpty( oldTranslations, mergedTranslationsCommented.new );
+
+                mergedTranslationsFileCommented = new File({
+                  path: namespacePathCommented,
+                  base: base,
+                  contents: new Buffer( JSON.stringify( mergedTranslationsCommented.new, null, 2 ) )
+                });
+                this.emit( 'writing', namespacePathCommented );
+                self.push( mergedTranslationsFileCommented );
+            }
         }
     }
 
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -2,7 +2,7 @@
 // turn it into a hash {foo: {bar: ""}}.
 // The generated hash can be attached to an
 // optional `hash`.
-function hashFromString(path, separator, hash) {
+function hashFromString(path, separator, hash, filePaths) {
     separator = separator || '.';
 
     if ( path.indexOf( separator, path.length - separator.length ) >= 0 ) {
@@ -14,11 +14,30 @@ function hashFromString(path, separator, hash) {
     var obj     = tmp_obj;
 
     for( var x = 0; x < parts.length; x++ ) {
-        if ( x == parts.length - 1 ) {
-            tmp_obj[parts[x]] = '';
+        if ( x === parts.length - 1 ) {
+            if (typeof filePaths !== 'undefined') {
+                tmp_obj[parts[x]] = {
+                    'msgstr': '',
+                    'paths': filePaths
+                };
+            } else {
+                tmp_obj[parts[x]] = '';
+            }
         }
         else if ( ! tmp_obj[parts[x]] ) {
-            tmp_obj[parts[x]] = {};
+            // prevent adding 'msgstr' and 'paths' to root
+            if (x === 0) {
+                tmp_obj[parts[x]] = {};
+            } else {
+                if (typeof filePaths !== 'undefined') {
+                    tmp_obj[parts[x]] = {
+                        'msgstr': '',
+                        'paths': filePaths
+                    };
+                } else {
+                    tmp_obj[parts[x]] = {};
+                }
+            }
         }
         tmp_obj = tmp_obj[parts[x]];
     }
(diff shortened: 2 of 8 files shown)


index.js
```diff
@@ -92,6 +93,15 @@ Parser.prototype._transform = function(file, encoding, done) {
 
 
 
+    // we find the relative path
+    // =========================
+    var relativeFilePath = path.relative(__dirname, this.base);
+    if (file.path) {
```
karellm: Is there a use case when `file.path` wouldn't be set?

index.js
```diff
@@ -189,21 +202,57 @@ Parser.prototype._flush = function(done) {
...
+        if (tempHash.hasOwnProperty(translation.key)) {
+            tempHash[translation.key] = tempHash[translation.key].concat(translation.paths);
+        } else {
+            tempHash[translation.key] = translation.paths;
+        }
 
+        // keep file paths in predictable order
+        tempHash[translation.key].sort();
 
+        return _;
```
karellm: Why do we need to return something here?

index.js
```diff
@@ -189,21 +202,57 @@ Parser.prototype._flush = function(done) {
...
+    // from
+    // [{ keyOne: '', paths: ['', '']}, { keyTwo: '', paths: ['']}]
+    // to
+    // { 'keyOne': { msgstr: '', paths: ['', ''] }, keyTwo: { msgstr: '', paths: [''] } }
+    // ==================================================================================
     for (var index in self.translations) {
         // simplify ${dot.separated.variables} into just their tails (${variables})
-        var key = self.translations[index].replace( /\$\{(?:[^.}]+\.)*([^}]+)\}/g, '\${$1}' );
+        var key = self.translations[index].key.replace( /\$\{(?:[^.}]+\.)*([^}]+)\}/g, '\${$1}' ),
+            paths = self.translations[index].paths;
```
karellm: Can you please use the following format:

```
var a;
var b;
```

src/helpers.js
```diff
@@ -14,11 +14,30 @@ function hashFromString(path, separator, hash) {
     var obj     = tmp_obj;
```
karellm: I don't think we need this line

src/helpers.js
```diff
@@ -36,13 +55,27 @@ function mergeHash(source, target, old, keepRemoved) {
...
     Object.keys(source).forEach(function (key) {
         if ( target[key] !== undefined ) {
-            if (typeof source[key] === 'object' && source[key].constructor !== Array) {
-                var nested = mergeHash( source[key], target[key], old[key], keepRemoved );
-                target[key] = nested.new;
-                old[key] = nested.old;
-            }
-            else {
-                target[key] = source[key];
+            if (typeof target[key] === 'object' && typeof target[key].msgstr !== 'undefined') {
```
karellm: This is exactly what I was afraid of, it is becoming impossible to maintain. The code complexity is way to high, which will make changing it painful in the future. My head hurts just trying to keep track of what's going on here. I think this must be broken down, either as a separate helper or into smaller chunks with better SRP.

index.js
sourceRS: In many of the tests, we're setting the 'base' of the fakeFiles we use, but not the path. This is why I initialized with this.base and conditionally override after that. So not a use case per say, but for the test cases, yes.

index.js
sourceRS: We don't necessarily need to, but I thought it was a little more explicit/clear that we're not modifying the translations while building the `tempHash`. Happy to remove it though, if you like :)

index.js
sourceRS: 👍

src/helpers.js
sourceRS: Actually I'm fairly certain we do, without it, tmp_obj gets overwritten with nested values so we wind up returning only part of the intended obj. 40 tests fail by removing this and returning tmp_obj instead.

index.js
karellm: I would prefer setting the `file.path` in the tests rather than having "test specific" code in the main script.

index.js
karellm: I tend not to return anything if we don't need it. I think it would be more consistent with the rest of the project.

src/helpers.js
karellm: Haha, yes we do :) My bad!

src/helpers.js
sourceRS: I would agree that the code is not very readable, but I'd have to disagree that what it's doing is complex. It's essentially just:

* Are we tracking paths?
 * yes, does the source contain path info?
    * yes, use the source's msgstr to set the target's msgstr
    * no, use source's value to set the target's msgstr
    * regardless, handle nested keys if they exist
 * no, handle old format as usual

One option would be to create two additional helpers to replace the conditionals and provide better readability, so it's more clear what is happening where:

```js
// function isPathObject () {...}
// function isNestedTranslation () {...}
// defined above^
if ( target[key] !== undefined ) {
    if ( isPathObject(target[key]) ) {
        // track paths
        if ( isPathObject(source[key]) ) {
            target[key].msgstr = source[key].msgstr;
        } else {
            target[key].msgstr = source[key];
        }

        if ( isNestedTranslation(source[key]) ) {
            var nested = mergeHash( source[key], target[key], old[key], keepRemoved );
            target[key] = nested.new;
            old[key] = nested.old;
        }
    } else {
        // don't track paths
        if ( isNestedTranslation(source[key]) ) {
            var nested = mergeHash( source[key], target[key], old[key], keepRemoved );
            target[key] = nested.new;
            old[key] = nested.old;
        }
        else {
            target[key] = source[key];
        }
    }
}
```

<hr/>

In addition, I could pull the handling of the new format out into a separate helper which gets called by `mergeHash`.

```js
// ... within mergeHash()
if ( target[key] !== undefined ) {
    if ( isPathObject(target[key]) ) {
        // track paths
        var hashWithPaths = mergeHashWithPaths( key, source, target, old, keepRemoved );
        target = hashWithPaths.target;
        old = hashWithPaths.old;
    }
    else {
        // don't track paths
        if ( isNestedTranslation(source[key]) ) {
            var nested = mergeHash( source[key], target[key], old[key], keepRemoved );
            target[key] = nested.new;
            old[key] = nested.old;
        }
        else {
            target[key] = source[key];
        }
    }
}
```

where `mergeHashWithPaths` is defined as:

```js
function mergeHashWithPaths(key, source, target, old, keepRemoved) {
    if ( isPathObject(source[key]) ) {
        target[key].msgstr = source[key].msgstr;
    }
    else {
        target[key].msgstr = source[key];
    }

    if ( isNestedTranslation(source[key]) ) {
        var nested = mergeHash( source[key], target[key], old[key], keepRemoved );
        target[key] = nested.new;
        old[key] = nested.old;
    }

    return {
        target,
        old
    }
}
```

Happy to do one or both of these, or try something else if you agree that it'll improve the readability/maintainability.

index.js
sourceRS: 👍

sourceRS:
Thanks for the review, @karellm!

So if I understand correctly, given a `translation.json` file in the original format of:

```json
{
  "key": "value",
  "key_plural": "values"
}
```

you're proposing either adding paths to the _plurals:

```json
{
  "key": {
    "msgstr": "value",
    "paths": ["..."]
  },
  "key_plural": {
    "msgstr": "values",
    "paths": ["..."]
  }
}
```

or nesting the _plurals inside the singular forms:

```json
{
  "key": {
    "msgstr": "value",
    "msgstr_plural": "values",
    "paths": ["..."]
  }
}
```

rather than leaving _plurals as-is, without path information:

```json
{
  "key": {
    "msgstr": "value",
    "paths": ["..."]
  },
  "key_plural": "values"
}
```

## tl;dr
**ratings out of 10 regarding the 'badness' of each downside is totally subjective.*
### adding paths to plurals, downsides:
1. (3/10 bad) duplicates information needlessly, resulting in a larger .po.json file.
2. (1/10 bad) is technically not an accurate representation of the strings in your project, since often times _plurals are added manually and do not correspond to a given file.

### nesting plurals in singular, downsides:
1. (10/10 bad) added complexity to _plural handling.
2. (10/10 bad) added complexity to gettext-converter's _plural handling.
3. (5/10 bad) less consistent with the original format (#of keys in .po.json !== #of keys in .json).

### conclusion: adding paths to plurals makes more sense.

## too long, read anyway
Nesting the _plurals in singular forms would require us to do more work to build the hash, as well as requiring additional changes to the handling of _plural forms in the gettext-converter. It'd also be less consistent with the original format, which seems like a generally bad idea even if there aren't any immediate consequences for breaking that consistency, especially in cases where the number of keys in the .json would not be identical to the number of keys in .po.json, just makes me nervous.

I'd say adding paths to the plural forms makes more sense, but I think there are some problems with that approach as well. Typically,  _plurals are not found in the source code, but are added manually - so there may not be a corresponding file path where a _plural form is found. This isn't an issue if we just accept that the _plural's file path is that of the singular form (which is very very likely to be true anyway), but it is an issue if people should expect the data to be clear and representative of reality (overall, not that big of a downside). Also, there's the issue of having multiple copies of the same data in different places, the .po.json is already quite a bit larger than the .json due to the objects replacing strings, so I'd think we don't want to make it any larger than it has to be.

Plus, the gettext-converter already knows how to locate _plural forms and place them at the correct line in the .po/.pot file under their singular forms. Perhaps I'm misunderstanding what the potential downsides are of allowing the `{msgstr: 'a value', paths: ['']}` format and `"a value"` format to live in the same file? The changes I've made to the gettext converter guarantee that it can read both formats simultaneously, but if the choice is between adding paths to plurals, or nesting plurals, I'd say adding paths to plurals makes a lot more sense, because it requires less complexity from both repos and the downside of extra bits in a .po.json file is a much smaller downside imo than added complexity to the source code.

Sorry haha I know this is getting dense, thanks again for the review!

sourceRS:
Everything's addressed and updated so far, aside from the .po.json formatting thing, going to go ahead and wait to address that until we clarify what the desired outcome is :) thanks again for the review, @karellm.

sourceRS:
Any preference from you on whether we nest the plurals inside the singular forms, put the plurals into the same, consistent, new format, or leave it as-is, @karellm?

sourceRS:
Hey @karellm I know the holidays are a busy time, just checking in again to see if you have a sec to clarify which direction you'd like to go with this :)

test/test.js
```diff
@@ -9,6 +9,15 @@ var hashFromString  = helpers.hashFromString;
...
 
+var emptyTranslationWithPaths = {
+    'msgstr': '',
+    'paths': ['']
+};
+
+// use with emptyTranslationWithPaths to prevent mutating the original
+function simpleDeepCopy(obj) {
+    return JSON.parse(JSON.stringify(obj));
+}
```
karellm: It looks like every single place use the same argument. Why not making a `emptyTranslationWithPaths()` method instead? It would simplify the code a little bit.

sourceRS:
Hey there @karellm, thanks for getting back to me! and my apologies for the headaches haha I know it's sort of a big addition. Really appreciate you taking the time to review it :)

I'm thinking maybe you meant 'hashFromString', rather than 'mergeHash', about adding an argument? Since I didn't change the arity of mergeHash in this PR.

To better separate the code from the base functionality, I could have a second function called 'hashFromStringWithPaths', which supports both formats. Then I could keep hashFromString unmodified and conditionally apply the correct function in the index file: [here](https://github.com/i18next/i18next-parser/pull/56/files#diff-168726dbe96b3ce427e7fedce31bb0bcR248), if that makes sense? I think originally I was just thinking that since we need to support the old format anyway while tracking the file paths (because of plurals and such), that rather than adding a new function to handle the new and old format simultaneously I could modify the function that already handles the old format to also support the new one. But if you'd rather have it separated I think that's totally reasonable as well.

I can also try and clean up the tests a bit, and move the track-paths tests into their own file, I'm definitely interested in getting this code merged in, even if it means making a few more tweaks until you're satisfied with the changes.

Happy Holidays, Happy New Year, and thanks again for reviewing all this! :)

test/test.js
sourceRS: yeah, you're right, this can be simplified.

sourceRS:
Split the tests out into a few separate files, and cleaned them up a bit, as well as rewording some things. Also updated hashFromString to remain as it was, and added hashFromStringWithPaths that implements the new version which handles either format, and that function is then used directly in index.js

Hopefully this is more along the lines of what you're looking for? and if not, I don't mind making more changes if it'll help this get merged in :)

libbymo:
@karellm Wanted to check on the status of this, given changes were made and wasn't sure if they had been reviewed. Is there anything we can do further to get this merged?

sourceRS:
https://github.com/i18next/i18next-gettext-converter/pull/56 should be good to go, but I believe @perrin4869 just wants to wait on this PR to get merged first, which makes sense.

karellm:
@sourceRS @libbymo Apologies for the delay. I've been busy with a project. I will get this merged but I also want to give it a proper review. I'm actually using this project again and plan on improving it soon. I want to take into consideration how this PR will integrate with everything else.

Thanks for your patience!

sourceRS:
okay, sounds good, thanks for the update! really appreciate it :)

libbymo:
@karellm Sounds good. Excited to hear your feedback. Thank you.

karellm:
I just released `1.0.0-beta1`. I will take a look at this next. I was curious if you were still using this and if there were some learnings you had in the process? Thanks

PR #63, opened 2017-05-09
Add test to parse function call on different line

johanneserber:
Adds a test for parsing constructs like
```js
const translatedToken = t(
  'some content'
);
```
Which now seems to be supported but nowhere documented.


karellm:
Thanks for the contribution!

PR #66, opened 2017-09-11
Add license to the package.json.

reiz:
I'm working on the open source project [VersionEye](https://www.versioneye.com). By adding the license info to the package.json it becomes available through the public NPMRegistry API and that way other tools like VersionEye can fetch it easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "Command Line tool for i18next",
   "name": "i18next-parser",
   "version": "0.12.0",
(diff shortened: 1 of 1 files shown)


karellm:
Thank you for the contribution! Didn't realized it would be that simple, I was expecting to have to copy some license file over.

reiz:
@karellm It's that simple :-) Now the license will appear on NPM and VersionEye.

PR #68, opened 2017-11-22
Add --default-values: JS default value extraction

jsomsanith:
i18next offers the ability to pass a default value within the code.
In translation world, the default value (in the default language) is crucial, as it helps to set a context, where only the key can be a technical key.

Extracting the keys is not sufficient in that case, we need to extract the default values too.

Add an option in cli for retro-compatibility `--default-values`.
When activated, it will parse the javascript code to get the default value from `i18n.t()` arguments, and add them in the generated json file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ node_modules/
 npm-debug.log
 
 # project files
+.idea/
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -34,6 +34,7 @@ program
   .option( '--keep-removed'                      , 'Prevent keys no longer found from being removed' )
   .option( '--write-old <string>'                , 'Save (or don\'t if false) _old files' )
   .option( '--ignore-variables'                  , 'Don\'t fail when a variable is found' )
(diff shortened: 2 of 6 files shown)


jsomsanith:
Plop, it's been 2 weeks, and no response on this PR. Can any maintainer react on it please ? :)

jsomsanith:
plop ? :)

nickasd:
It's a pity this PR has not been merged yet. I'm looking for a way to include default values in the code as well. Not only does this default value act as a sort of documentation for the key, it is also easier for translators to see if the string contains formatters (like `{}` or `%s`) and provide a valid translation.

Besides, I would suggest that the first comma in the regular expression be optional, so that the default value can be passed as an additional argument in .html files when using Handlebars. An example would be: `{{t key default}}`.

karellm:
This was merged (manually since it introduced tabs when this project uses spaces). Thanks for the contribution and patience. I published `0.13.0` with the feature.

jsomsanith:
Thank you @nickasd and @karellm

PR #69, opened 2017-11-28
Add --react-namespace: parse namespace from react component

GitBalou:
i18next-parser provide the ability to extract namespace from `t` function when provided with the key : `t('namespace:key')`
However, when using react-i18next and the high order component `translate`, the namespace is not provided with the `t` function but with the translate hoc.
```
class Component extends React.Component {
  ...
 // somewhere in jsx
  <p>{t('key')}</p>
}

// export component with translate
export default translate('namespace')(component)
```

This PR provide a new option `--react-namespace` which default to false. If set to true, i18n-parser will look for `translate` hig order component and use parsed namespace instead of `self.defaultNamespace`.


karellm:
From the test it is not clear what this adds to what can already be done. The translation functions syntax is the same as any JS code. Can you explain?

GitBalou:
When using react with i18next, you can create a high order component with `translate('componentNamespace')(component)`. Then in your component JSX you can do something like `<p>{t('hello')}</p>`. i18next will set namespace to `componentNamespace` as defined in translate HOC.
However, i18next-parser will set namespace to `self.defaultNamespace`, as there is no namespace defined in `t` function.

PR #70, opened 2017-11-30
Add parsing for namespace from react high order component Translate

GitBalou:
i18next-parser provide the ability to extract namespace from `t` function when provided with the key : `t('namespace:key')`
However, when using react-i18next and the high order component `translate`, the namespace is not provided with the `t` function but with the translate hoc.
```
class Component extends React.Component {
  ...
 // somewhere in jsx
  <p>{t('key')}</p>
}

// export component with translate
export default translate('namespace')(component)
```

This PR provide a new option `--react-namespace` which default to false. If set to true, i18n-parser will look for `translate` hig order component and use parsed namespace instead of `self.defaultNamespace`.


Godefroy:
We need this at Lone Stone, could you please merge this branch ? :-)

karellm:
I just published `1.0.0-beta2` which adds support for react. I invite you to test it and open a new ticket if there are issues.

PR #71, opened 2018-01-25
Add support for the $LOCALE and $FILENAME tags in the namespace

nickasd:
When a project is structured such that single files represent a namespace each, it is not desirable to include the namespace for each key in the source code. With this addition, it will be possible to inject the $FILENAME tag in the namespace such that each source file can have its own translation file.

With this addition, the --prefix and --suffix options are superfluous since they can be also added to the namespace. They are still supported for backward compatibility but not mentioned anymore in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,9 +71,9 @@ Thanks a lot to all the previous [contributors](https://github.com/i18next/i18ne
 - **--keep-removed**: Prevent keys no longer found from being removed
 - **--write-old false**: Avoid saving the \_old files
 - **--ignore-variables**: Don't fail when a variable is found
-- **--prefix <string>**: Prefix filename for each locale, eg.: 'pre-$LOCALE-' will yield 'pre-en-default.json'
-- **--suffix <string>**: Suffix filename for each locale, eg.: '-$suf-LOCALE' will yield 'default-suf-en.json'
-- **--extension <string>**: Specify extension for filename for each locale, eg.: '.$LOCALE.i18n' will yield 'default.en.i18n'
+- **--extension <string>**: Edit the extension of the files. Defaults to `.json`
+
+You can inject the locale tag in the namespace and extension using the `$LOCALE` variable, and you can inject the source file name in the namespace using the `$FILENAME` variable.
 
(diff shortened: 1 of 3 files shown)


karellm:
@nickasd Thanks for the PR. After a quick glance, this does not look backward compatible. It was supporting `prefix` and `suffix` and now only support a prefix via an obscure API. It does not look possible to add both a prefix and a suffix with your changes. 

I'm open to the `$FILENAME` idea but please integrate that inside the existing `prefix` and `suffix` options, not as a replacement.

nickasd:
Thanks for taking the time to review it.
You're right, there is a small bug which causes only the first appearance of the $LOCALE tag to be replaced. Is this what you were referring to? It was also my mistake to remove the test for the --prefix and --suffix options, since they should still test for backward compatibility.

In my changes you can add a prefix and/or a suffix inside the namespace option. I think this format is more versatile since the user can specify everything in one string/option and the code doesn't need to check namespace, prefix, suffix and extension for tags to be replaced. Do you see what I mean? If you still think this is not good, I can restore the previous behaviour.

karellm:
I read your code to quickly, you are right that the namespace alone is necessary.

- [ ] Keep old tests as long as the feature is supported
- [ ] Add deprecation warning if they are used
- [ ] Make sure many `$LOCALE`, `$FILENAME` are replaced.

As a FYI, I'm currently rewriting this package entirely with newer ES6 syntax and all the learning from the many use cases that popped over the years and modified the code into a mess. I expect a beta to land next week.

nickasd:
I've corrected those issues, thanks for the suggestions.

Looking forward to the new release!

karellm:
As I was working on #72 I realized this is not as easy as it looks. One catalog contains keys from many files normally. I don't think that `$FILENAME` belongs into the code. If you run it one file at a time, then change the `filename` option there directly.

I will close this but let me know if I missed something.

nickasd:
Thanks for letting me know. I ended up not using `$FILENAME` myself, but having only one `filename` option looks good.

Issue #67, opened 2017-10-24
Problem with : (colon) inside text


burgerboydaddy:
Hello,
I'm trying to parse text with : (colon) inside it.
For example: "Hello, I have: one, two, three"
Parser will separate this into 2 strings. "Hello, I have" and "one, two, three". 
Any suggestion how to handle this case (without splitting into separated strings)?
Thanks

karellm:
It looks like a bug. Can you make a PR with a failing test case? Thanks!

opposite-bracket:
@burgerboydaddy I had the same issue but this can be solved by defining a different `namespaceSeparator`. Here's an example of what my function call in gulp looks like:

```
gulp.task('i18next:parser', function () {
...
      .pipe(i18nextParser({
          ...
          namespaceSeparator: '__SS',
          ...
      }))
...
});
```

Of ocurse, we're not using namespaces but it is up to you to configure it as you want it. Hope that helps.

karellm:
This will also be addressed in `1.0.0`. You can follow progress in #72. It should be ready with a week or two.

karellm:
I second what @opposite-bracket suggested. After looking into it, this is not a bug but a feature. If you rely on keys that are sentences and might use `:`, you should disable or change the `namespaceSeparator` option.

Issue #61, opened 2017-04-27
yaml output?


bonesoul:
i'm using https://github.com/i18next/i18next-node-fs-backend with yaml files, is it possible for i18next-parser to output yaml files too instead for json?

karellm:
@bonesoul It is possible but if you want it soon, you should make a PR for it.

karellm:
This is addressed in #72. In `1.x`, set `extension` to `.yml` to get a yml output.

Issue #57, opened 2016-11-30
Doesn't extract keys from Interpolation component with react-i18next


isld:
Can't be much more descriptive than in the title. Here is a code snippet:

```
<div className="flex-item-m-2 recent-activity-item-footer">
  <Interpolate i18nKey='Total {{amount}}' useDangerouslySetInnerHTML={true} amount={totalDocs} />
</div>
```

Sadly, the string 'Total {{amount}}' never makes it to the JSON output of the parser.

karellm:
@isld I looked into this. It wouldn't be too hard to add but I'm somewhat reluctant to add more trackers to the code. It might be time to have a better configuration to support different needs (people using react likely don't care about html parsing or handlebars). Would you be interested in submitting a PR?

hotdang-ca:
I may take this on as well...

davelandry:
I was able to override the default regex to grab both <Interpolate> components and translate functions. It's rudimentary, and assumes the codebase enforces double-quotations, but it should be a start:
```sh
-p "i18nKey=\"([^\"]+)\"|t\(\"([^\"]+)\""
```

tkrotoff:
The problem is more complex, you have multiple cases in order to support react-i18next:
```JSX
<Trans i18nKey="hello">Hello <strong>World!</strong></Trans>
<Trans>Hello, <strong>World!</strong></Trans>
```

```JSON
{
  "hello": "Bonjour <1>le monde !</1>",
  "Hello, <1>World!</1>": "Bonjour <1>le monde !</1>"
}
```

karellm:
I will write a React Lexer for the `1.x` release. I'm currently working on matching existing features and should land a beta this week.

karellm:
I just published `1.0.0-beta2` which adds support for react. I invite you to test it and open a new ticket if there are issues.

Issue #55, opened 2016-10-12
Tracking file path references for each key


sourceRS:
## tldr:

Wanting to add the functionality to track file path references to each key for the purpose of having those comments end up in .po files for our translators to get better context as-needed, requiring modifications both in this repo and in the [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter), and curious how amenable you'd be to merging that sort of thing in?

---

## long form:

I've been helping with some efforts to internationalize a medium-large size Ember app, and since [gettext](https://www.gnu.org/software/gettext/) doesn't play nice with html and handlebars, we needed these two repos to generate our translation strings and convert between .json -> .po and vice versa.

But we ran into a slight issue, our translators will need file path references for each key, as comments in the .po files (something that gettext does by default in a typical i18n workflow, except gettext has line numbers, which we won't be needing) and although the gettext-converter does have some* ability to add comments, the data structure output by this tool doesn't keep track of those file paths.

Initially I modified the i18next-parser to output something that looks more like: 

```
{
  'some key': {
    msgstr: '',
    paths: ['app/templates/error.hbs', '../fake/path']
  }
}
```

Where `paths` is an array of each file path location that a given key is found at, in case you have the same key in multiple places in your source code.

rather than the current output:

```
{
  'some key': ''
}
```

and I modified the gettext-converter to handle both formats dynamically while converting from .json -> .po, so you can have keys with file path information and others without, in the same file.

Currently, my little demo of this working involves either symlinking my local changes or pointing our app at my own forks of these two repos, which is less than ideal. I'm hoping to get these changes merged in at some point, so in the past couple days I've gone through and modified my code to use an additional flag '-t, --track-paths', so it generates the original data structure unless you include that flag and set it to true.

To further support our workflow, since the i18next runtime still needs the original format, I also made it so that using '--track-paths' will prepend '/tmp' to the output location of the translation file but not the translation_old file. So rather than running this tool to get our .json, and then running the converter to get our .po - we are running this tool with `--track-paths true` to get our new .json format, then running the converter on the output /tmp/locales/en/translation.json to get our .po file, and finally running the converter a second time to go from .po -> .json so that we're back in the original format, and now back in the correct directory as well.

[...]

PR #72, opened 2018-01-26
1.0.0

karellm:
## Roadmap for the release:

- [x] Rewrite helpers
- [x] Write a Lexer per supported language
  - [x] HTML
  - [x] Handlebars
  - [x] JS
- [x] Write a parser that auto-detect the file extension and call the proper Lexer
- [x] Update CLI code
  - [x] Transition to `vinyl-fs`
  - [x] Compile ES6 code
  - [x] Test cli
- [x] Bugs and other
  - [x] Make sorting an option #74 
  - [x] Make indentation an option #75 
  - [x] Remove `gulp-util`
  - [x] Update lexers options
  - [x] Fix defaultValue bug #43 
  - [x] colon in key #67
  - [x] Better error handling ([PluginError](https://github.com/gulpjs/plugin-error)?)
  - [x] YAML output #61 
  - [x] Create context keys ([see this](https://github.com/i18next/i18next-parser/blob/master/index.js#L124)) 
- [x] Update the documentation
  - [x] CLI
  - [x] Gulp
  - [x] Write transition document

## Roadmap post release:

- [ ] Track file path #55 
- [ ] Write a Lexer for JSX #57 
- [ ] Add documentation Grunt/Webpack
- [ ] Handle [arrays/objects](https://www.i18next.com/objects-and-arrays.html) in catalog
- [ ] `useKeyAsDefaultValue`

## Other

- [x] ~[Prettier](https://github.com/prettier/prettier)~
- [ ] [Coveralls](https://coveralls.io/)


birdy-:
Since you rewrite many things, take a look at prettier. It's a great tool to clean up code. Just add `.prettierrc` and use their binary or VSCode/Atom plugin.

karellm:
@birdy- thanks I will definitely look at it. My code style changed quite a bit in the last 4 years, every time I had to maintain that code base was an humbling moment 😄 The new structure will hopefully make that a lot more pleasant. 

As to opinionated code styles, they make sense until they don't and then get in the way. That said I definitely should configure a linter of sort for collaborators to know what rules to follow.

tkrotoff:
I have a suggestion: generating JS files (e.g `locales/fr/translation.js`). We would then have comments:
```JS
{
  // Extracted from file src/utils/Colors.js
  "Red": "Rouge",
  "Orange": "Orange",
  "Yellow": "Jaune",
  "Green": "Vert",
  "Blue": "Bleu",
  "Indigo": "Indigo",
  "Violet": "Violet",

  // Extracted from file src/utils/Languages.js
  "English": "Anglais",
  "French": "Français",

  // Extracted from file src/SignUpForm.js
  "First Name": "Prénom",
  "Last Name": "Nom de famille",
  "Email": "Email"
}
```

It's nice for the translator to have the "context", like [gettext's PO files](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html):
```PO
#: lib/error.c:116
msgid "Unknown system error"
msgstr "Error desconegut del sistema"
```

karellm:
@tkrotoff My understanding is that #55 would address that. Is there something more you'd want?

tkrotoff:
@karellm thx for the link. #55 proposes to use JSON, that's better than my proposition that relies on JS comments: you can then easily extract `paths` with a tool if needed.

birdy-:
Prettier will format code with few option (max length of a line, simple/double quote, ...). It's great because you don't have to do it manually.
Prettier is not a `eslint --fix` because it won't change anything in your code. It means, if you use a `let` instead of a `const` or do some useless import, it won't change it.
So prettier is not "enough", you have to use a lint. Since `airbnb-eslint` is not "compatible" with prettier out of the box, you can extend your rules with `airbnb` and `prettier` to solve incompatibilities. https://github.com/prettier/prettier-eslint

If you need help or you want me to setup it, I can do it.

karellm:
It's almost there. `1.0.0-beta1` is coming very soon!

karellm:
@birdy- I won'd use prettier because it transforms:

```
if() {
}
else {
}
```

into

```
if() {
} else {
}
``` 

which makes absolutely no sense to me.

birdy-:
> By far the biggest reason for adopting Prettier is to stop all the on-going debates over styles.
:D

karellm:
Well there is something more important than rigidity. It's consistency, readability and actual productivity while coding. Prettier doesn't score well in 2 of the 3 for me.

PR #73, opened 2018-02-03
Fix tests on Windows

TheDancingCode:
This PR normalizes file paths in the tests to make the tests pass on Windows.

### DIFF
--- a/test/parser.js
+++ b/test/parser.js
@@ -7,7 +7,7 @@ describe('parser', function () {
         });
 
         i18nextParser.on('data', function (file) {
-            if ( file.relative === 'en/translation.json' ) {
+            if ( file.relative === path.normalize('en/translation.json') ) {
                 result = JSON.parse( file.contents );
             }
         });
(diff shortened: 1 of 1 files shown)


karellm:
Hi, I'm a little confused by this PR.`path.normalize('en/test_plural.json')` will return exactly `en/test_plural.json`. Can you describe the issue you're having and the solution you found? 

I'm closing this, not by lack of interest but because `master` changed so much that a new PR might make more sense at this point.

TheDancingCode:
Paths on Windows have a different segment separator: `\` instead of `/`. Because of this, expressions like `file.relative === 'en/test_plural.json'` will return `false`, since `file.relative` equals `'en\test_plural.json'` on Windows. A simple fix is to normalize the path; an operation that will indeed return `'en/test_plural.json'` on your system, but `'en\test_plural.json'` on Windows systems, like mine. Another way to achieve this is to use `path.join('en', 'test_plural.json')`.

This is fixing a problem that you don't have, but that Windows users face.

karellm:
Thanks for the clarification. I will fix the test on master branch since `1.0.0-beta` changed the code base a lot!

karellm:
@TheDancingCode I fixed them but have no way to test. I would appreciate if you could have a look: https://github.com/i18next/i18next-parser/commit/96a49aa0e098cad995c75049564f57ebbc1ada35

Thanks for your contribution!

PR #75, opened 2018-02-08
Allow indentation level to be specified as a cli parameter

lucsky:
This PR allows to specify the indentation level of the final output content from a CLI parameter, as so:

`i18next /path/to/file/or/dir --indentation 4 -o /output/directory`


karellm:
This is addressed in `1.x`. You can track progress at #72. Closing this for now as I won't add it as part of `0.x`.

PR #82, opened 2018-03-24
Fix react test to have valid jsx

coyotte508:
`Trans` and `Interpolate` need to be imported to be used.

By the way, i18next-parser should not extract the whole html tags, but use <0>, <1> and so on as seen [here](https://react.i18next.com/components/trans-component.html).


karellm:
Thanks but it doesn't matter much. This code is never executed, it is only here to be parsed for tests.

As to the `<0>` and `<1>`, I'm happy to welcome a PR. I currently don't have the time to implement this.

PR #85, opened 2018-04-02
Improved parsing of Trans component

coyotte508:
Based on https://react.i18next.com/components/trans-component.html

Each html tag (closing or self-closing) and each js expression is converted to `<x>content</x>` where `x` is the index of tag/expression in the array of children of the current tag (`Trans` being the root tag).

The example:

```jsx
Hello <strong title={t('fourth')}>{{name}}</strong>, you have {{count}} unread message. <Link to="/msgs">Go to messages</Link>.
```
Becomes:

```jsx
Hello <1><0>{{name}}</0></1>, you have <3>{{count}}</3> unread message. <5>Go to messages</5>.
```

You may notice that I deviated from the usual way of parsing. I use regexps for parsing quotes, but for the rest I don't use regexps. This avoids some caveats with regexps, for example `<a stuff="a>" ....>` with regexps the first closing `>` would trigger the end of the tag even though it shouldn't as it's surrounded by quotes.

The js parser *could* be improved to try and detect keys of objects, for example `{{name: 'Albert'}}` is now parsed as is, but it should render  `{{name}}`. That said, it's good practice to set the variables before rendering the jsx, and as long as the user of the parser is aware, there should be no problems.

### DIFF
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -1,4 +1,7 @@
+import assert from 'assert'
 import HTMLLexer from './html-lexer'
+import BaseLexer from './base-lexer';
+import { ParsingError } from '../helpers';
 
 export default class JsxLexer extends HTMLLexer {
   constructor(options = {}) {
@@ -44,7 +47,7 @@ export default class JsxLexer extends HTMLLexer {
       const key = attrs.keys
 
       if (matches[3] && !attrs.options.defaultValue) {
-        attrs.options.defaultValue = matches[3].trim()
+        attrs.options.defaultValue = this.eraseTags(matches[3].trim()).replace(/\\s+/g, ' ')
       }
 
       if (key) {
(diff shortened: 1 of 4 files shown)


karellm:
Thanks for the PR, I will review it later today or tomorrow

coyotte508:
Alright, let me know if there's anything I can do to make it easier :)

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +57,168 @@ export default class JsxLexer extends HTMLLexer {
...
+      }
+    }).join('')
+  }
+
+  /**
+   * Convert jsx into an array of children
+   * @param {*} string
+   */
+  parseJsx(string) {
+    if (string.length === 0) {
```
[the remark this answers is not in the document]
coyotte508: The problem is that it would not parse `'0'` correctly in that case. As `'0' == false.

coyotte508:
The main cons of the current approach is that I require all tags to be closed, e.g. use of `<br />` instead of `<br>`. 

Regarding other parsers, neither https://www.npmjs.com/package/react-html-parser nor https://www.npmjs.com/package/html-react-parser parse embedded javascript in html. And they require react as a dependency.

Basing the parser on `htmlparser2` would not be perfect either, as it doesn't parse javascript and could be broken with some custom jsx.

I looked at https://www.npmjs.com/package/jsx-parser , and it's not great either (breaks on the examples, no gulping spaces).

I see two possibilities: 
- Improve the current code so that tags like `<br>`, `` and `<input>` don't need a closing tag, add a lot of tests, etc, maybe make the jsx parser into a whole separate module.
- Use html2parser or parse5, parse text elements to detect embedded javascript. Good solution for most cases but can never handle everything. (eg `<a> abbc {'</a>'} azdc </a>` will always break). Minimal code addition / code to review.

To be honest I don't care which of the two solutions is used, I just really need the functionality added to `i18next-parser`.  Tell me which solution you prefer, and I'll go for it.

coyotte508:
@karellm It's done, using acorn-jsx: https://github.com/RReverser/acorn-jsx.

None of the other parsers I found, including react-html-parser and html-react-parser, htmlparser2, parse5, could handle js expressions mixed in html.

src/lexers/jsx-lexer.js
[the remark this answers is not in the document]
karellm: Well `!'0'` is `false` and `'0'.length === 0` is also `false`.

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +58,59 @@ export default class JsxLexer extends HTMLLexer {
...
+    return elemsToString(children);
+  }
+
+  /**
+   * Simplify the bulky AST given by Acorn
+   * @param {*} children An array of elements contained inside an html tag
+   * @param {string} originalString The original string being parsed
+   */
+  simplify(children, originalString) {
+    for (let i = 0; i < children.length; i ++) {
```
karellm: Is there a reason you don't use `forEach` here? If not, it would be more consistent with the rest of the code

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +58,59 @@ export default class JsxLexer extends HTMLLexer {
...
+
+  /**
+   * Recursively convert html tags and js injections to tags with the child index in it
+   *
+   * @param {string} string
+   *
+   * @returns string
+   */
+  eraseTags(string) {
+    const children = this.simplify(acorn.parse(string, {plugins: {jsx: true}}).body[0].expression.children, string);
```
karellm: Please avoid line that are longer than 80 char. I would suggest doing this in two steps

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +58,59 @@ export default class JsxLexer extends HTMLLexer {
...
+
+    const elemsToString = children => children.map((child, index) => {
```
karellm: Is there a reason you create a function rather than directly returning the mapped children?

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +58,59 @@ export default class JsxLexer extends HTMLLexer {
...
+
+    return elemsToString(children);
+  }
+
+  /**
+   * Simplify the bulky AST given by Acorn
+   * @param {*} children An array of elements contained inside an html tag
+   * @param {string} originalString The original string being parsed
+   */
+  simplify(children, originalString) {
```
karellm: Can you name this something a little more idiomatic like `parseAcornPayload`?

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +58,59 @@ export default class JsxLexer extends HTMLLexer {
...
+          content: originalString.slice(child.start, child.end)
+        }; break;
+        default: throw new ParsingError("Unknown ast element when parsing jsx: " + child.type)
+      }
+    }
+
+    // Filter empty text elements. Using string.length instead of !string because
+    // '0' is a valid text element, and '' is not, and !string doesn't make a difference
+    // between the two.
+    children = children.filter(child => !(child.type === 'text' && child.content.length === 0));
```
karellm: This could be written `child.type !== 'text' || child.content` which is easier to read imo

I'm not sure about your comment either. As I commented above: `'0'` is truthy, `''` is falsy. They are not the same.

src/lexers/jsx-lexer.js
coyotte508: Yes, it calls itself recurisvely.

src/lexers/jsx-lexer.js
coyotte508: Right, `'0' == false`, yet `!'0' == false` too... A quirk of js

coyotte508:
Thank you for your fast review. I just pushed some changes, don't hesitate to take over if needed.

src/lexers/jsx-lexer.js
```diff
@@ -54,4 +57,168 @@ export default class JsxLexer extends HTMLLexer {
...
   }
+
+  /**
+   * Recursively convert html tags and js injections to tags with the child index in it
+   *
+   * @param {string} string
+   *
+   * @returns string
+   */
+  eraseTags(string) {
```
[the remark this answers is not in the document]
coyotte508: I added a test for `eraseTags`.

karellm:
Thanks for the contribution, `1.0.0-beta9` was just released

PR #90, opened 2018-05-01
added support for keyless react-i18next Trans components

MynockSpit:
react-i18next <Trans> components have support for using the content as the key. This adds support for that.

Note that we probably should move away from regex and towards an actual parser of some kind in the future... this (and the previous regex) will break if someone has an attribute on their Trans component with a '>' in it.

### DIFF
--- a/dist/lexers/jsx-lexer.js
+++ b/dist/lexers/jsx-lexer.js
@@ -32,20 +32,25 @@ JsxLexer = function (_HTMLLexer) {_inherits(JsxLexer, _HTMLLexer);
       }
 
       return this.keys;
-    } }, { key: 'extractTrans', value: function extractTrans(
+    }
 
+    /**
+      * Extract tags and content from the Trans component.
+      * @param {string} string
+      * @returns {array} Array of key options
+      */ }, { key: 'extractTrans', value: function extractTrans(
     content) {
       var matches = void 0;
-      var closingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)\\/>)';
-      var selfClosingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)>((?:\\s|.)*?)<\\/Trans>)';
+      var selfClosingTagPattern = '(?:<\\s*Trans([^>]*)?/>)';
+      var closingTagPattern = '(?:<\\s*Trans([^>]*)?>((?:(?!</\\s*Trans\\s*>)[^])*)</\\s*Trans\\s*>)';
       var regex = new RegExp(
-      [closingTagPattern, selfClosingTagPattern].join('|'),
+      [selfClosingTagPattern, closingTagPattern].join('|'),
       'gi');
 
 
       while (matches = regex.exec(content)) {
         var attrs = this.parseAttributes(matches[1] || matches[2]);
-        var key = attrs.keys;
+        var key = attrs.keys || matches[3];
 
         if (matches[3] && !attrs.options.defaultValue) {
           attrs.options.defaultValue = this.eraseTags(matches[0]).replace(/\s+/g, ' ');
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -34,18 +34,23 @@ export default class JsxLexer extends HTMLLexer {
     return this.keys
   }
 
+  /**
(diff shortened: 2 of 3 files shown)


src/lexers/jsx-lexer.js
```diff
@@ -34,18 +34,23 @@ export default class JsxLexer extends HTMLLexer {
...
+  * Extract tags and content from the Trans component.
+  * @param {string} string
+  * @returns {array} Array of key options
+  */
   extractTrans(content) {
     let matches
-    const closingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)\\/>)'
-    const selfClosingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)>((?:\\s|.)*?)<\\/Trans>)'
+    const selfClosingTagPattern = '(?:<\\s*Trans([^>]*)?/>)'
+    const closingTagPattern = '(?:<\\s*Trans([^>]*)?>((?:(?!</\\s*Trans\\s*>)[^])*)</\\s*Trans\\s*>)'
```
MynockSpit: 1. The `\\s*` are places where any amount of space can syntactically occur.
2. Removed strict checking for `this.attr` b/c parseAttributes does that already.
3. Renamed `closingTagPattern` to `selfClosingTagPattern` and vice versa (they were backwards.)

dist/lexers/jsx-lexer.js
```diff
@@ -32,20 +32,25 @@ JsxLexer = function (_HTMLLexer) {_inherits(JsxLexer, _HTMLLexer);
...
       var regex = new RegExp(
-      [closingTagPattern, selfClosingTagPattern].join('|'),
+      [selfClosingTagPattern, closingTagPattern].join('|'),
       'gi');
 
 
       while (matches = regex.exec(content)) {
         var attrs = this.parseAttributes(matches[1] || matches[2]);
-        var key = attrs.keys;
+        var key = attrs.keys || matches[3];
```
MynockSpit: If we didn't get a key back from `parseAttributes`, use the content we matched (which is `undefined` if we didn't match it or `''` if we matched nothing -- both of which are falsey.)

src/lexers/jsx-lexer.js
```diff
@@ -34,18 +34,23 @@ export default class JsxLexer extends HTMLLexer {
...
     const regex = new RegExp(
-      [closingTagPattern, selfClosingTagPattern].join('|'),
+      [selfClosingTagPattern, closingTagPattern].join('|'),
       'gi'
     )
 
     while (matches = regex.exec(content)) {
       const attrs = this.parseAttributes(matches[1] || matches[2])
-      const key = attrs.keys
+      const key = attrs.keys || matches[3]
```
MynockSpit: If we didn't get a key back from `parseAttributes`, use the content we matched (which is `undefined` if we didn't match it or `''` if we matched nothing -- both of which are falsey.)

src/lexers/jsx-lexer.js
```diff
@@ -34,18 +34,23 @@ export default class JsxLexer extends HTMLLexer {
...
+  */
   extractTrans(content) {
     let matches
-    const closingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)\\/>)'
-    const selfClosingTagPattern = '(?:<Trans([^>]*\\s' + this.attr + '[^>]*?)>((?:\\s|.)*?)<\\/Trans>)'
+    const selfClosingTagPattern = '(?:<\\s*Trans([^>]*)?/>)'
+    const closingTagPattern = '(?:<\\s*Trans([^>]*)?>((?:(?!</\\s*Trans\\s*>)[^])*)</\\s*Trans\\s*>)'
     const regex = new RegExp(
-      [closingTagPattern, selfClosingTagPattern].join('|'),
+      [selfClosingTagPattern, closingTagPattern].join('|'),
```
karellm: Just curious if there is a reason you switched the order of the variables here and at lines 44?

MynockSpit: They were already in this order, just mislabeled. The 'closingTagPattern' was actually the 'selfClosingTagPattern' and vice versa. 

I changed the *perceived* order of them to the actual order of them. If I didn't change that, I'd have to reorder the matches later, and I was being lazy.

PR #92, opened 2018-05-02
Move JS to acorn parsing

karellm:
- [x] Move JS parsing to acorn
- [ ] Move React parsing to acorn

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Option         | Description                 | Default
 Option        | Description                 | Default
 ------------- | --------------------------- | -------
 **functions** | Array of functions to match | `['t']`
+**acorn**     | Options to pass to acorn    | `{}`
 
 **`JsxLexer` options**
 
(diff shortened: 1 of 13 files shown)


karellm:
@MynockSpit Would you mind giving me a hand here? I'm not personally using react and did all the work for JS. Moving to the `acorn-jsx` parser per your suggestion would be a nice addition.

MynockSpit:
Sure, I can take a look.

MynockSpit:
I think it makes sense to combine the JS and JSX parsers -- the JSX parser is essentially an optional path on the JS parser. (If we're parsing a .jsx file without JSX, the JSX path will never get hit, and if we parse a .js file with JSX, the JSX path will get hit. It'll parse it correctly either way.)

Thoughts?

karellm:
@MynockSpit I tend to prefer two lexers in cases there would be differences later on. But I'm totally ok the React lexer to inherit the Javascript one instead of the HTML one.

I trust your judgement though as you dive into `acorn-jsx`. If it works seamlessly and it doesn't add too much complexity it might be ok to combine them.

MynockSpit:
Cool. I'll make a PR with them merged and comments (and pros/cons) and see what you think.

MynockSpit:
Let me know what you think: https://github.com/i18next/i18next-parser/pull/93

PR #93, opened 2018-05-03
moved JSX parsing to acorn-jsx (is now a side-effect of JS parsing)

MynockSpit:
Major notes:

1. JSX parsing now happens as a side effect of a JSX node existing in any JS file. 
    - The only major **pro** here is that jsx code inside a .js file gets parsed correctly instead of silently failing -- we also no longer need to ask the user to define a config param to pull from JSX. If the file is only js, the JSX paths never get hit.
    - Conversely, the only major **con** here is that the js lexer is a bit more complex. I kept as much as was easy out of the js lexer, but more could be moved. For example: the traps object could be dynamically generated -- if it was, we could move the JSXElement trap into the jsx-lexer file.
2. Changed the JS trap from `Expression` to `CallExpression` -- the JS was checking for it anyways in the body of the trap and CallExpression catches js nested in JSX. Was there a reason to favor `Expression` over `CallExpression`
3. Added a couple tests for user-defined key attributes.

(comments incoming)

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -1,14 +1,16 @@
-import * as acorn from 'acorn'
+import * as acorn from 'acorn-jsx'
 import * as walk from 'acorn/dist/walk'
 import { ParsingError } from '../helpers'
 import BaseLexer from './base-lexer'
+import { JSXParserExtension, nodeToString } from './jsx-lexer'
 
 export default class JavascriptLexer extends BaseLexer {
   constructor(options = {}) {
     super(options)
 
-    this.acornOptions = { sourceType: 'module', ...options.acorn }
+    this.acornOptions = { sourceType: 'module', plugins: { jsx: true }, ...options.acorn }
     this.functions = options.functions || ['t']
+    this.attr = options.attr || 'i18nKey'
   }
 
   extract(content) {
@@ -17,10 +19,9 @@ export default class JavascriptLexer extends BaseLexer {
     walk.simple(
       acorn.parse(content, this.acornOptions),
       {
-        Expression(node) {
+        CallExpression(node) {
           let entry = {}
           const isTranslationFunction = (
-            node.type === 'CallExpression' &&
             node.callee && (
               that.functions.includes(node.callee.name) ||
               node.callee.property && that.functions.includes(node.callee.property.name)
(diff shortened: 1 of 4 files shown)


src/lexers/javascript-lexer.js
```diff
@@ -1,14 +1,16 @@
-import * as acorn from 'acorn'
+import * as acorn from 'acorn-jsx'
```
MynockSpit: `acorn-jsx` imports `acorn`, adds to it, and exports it again.

src/lexers/javascript-lexer.js
```diff
@@ -17,10 +19,9 @@ export default class JavascriptLexer extends BaseLexer {
     walk.simple(
       acorn.parse(content, this.acornOptions),
       {
-        Expression(node) {
+        CallExpression(node) {
```
MynockSpit: Two reasons:
1. We were already checking if it were a CallExpression, and not doing anything on any other types.
2. Setting it to CallExpression lets it work in both js and jsx syntax.

src/lexers/javascript-lexer.js
```diff
@@ -57,8 +58,45 @@ export default class JavascriptLexer extends BaseLexer {
 
             that.keys.push(entry)
           }
+        },
+        JSXElement(node) {
```
MynockSpit: This is the trap for JSX Elements.

Check if it's "Trans" --> Set the key --> Set the defaultContent --> If the set key is falsey and there is defaultContent, set the key to defaultContent --> if we now have a key, push to the keys array.

Check if it's "Interpolate" --> Set the key --> if we have a key, push to the keys array.

src/lexers/jsx-lexer.js
```diff
@@ -1,120 +1,69 @@
...
-          type: 'text',
-          content: child.value.replace(/^(?:\s*(\n|\r)\s*)?(.*)(?:\s*(\n|\r)\s*)?$/, '$2')
-        }
-      }
-      else if (child.type === 'JSXElement') {
-        return {
-          type: 'tag',
-          children: this.parseAcornPayload(child.children, originalString)
-        }
+function parseAcornPayload(children, originalString) {
```
MynockSpit: Despite the diff, there's no change here except for moving it out of the class.

src/lexers/jsx-lexer.js
```diff
@@ -1,120 +1,69 @@
...
-  */
-  extractTrans(content) {
-    let matches
-    const selfClosingTagPattern = '(?:<\\s*Trans([^>]*)?/>)'
-    const closingTagPattern = '(?:<\\s*Trans([^>]*)?>((?:(?!</\\s*Trans\\s*>)[^])*)</\\s*Trans\\s*>)'
-    const regex = new RegExp(
-      [selfClosingTagPattern, closingTagPattern].join('|'),
-      'gi'
-    )
+export function nodeToString(ast, string) {
```
MynockSpit: This is identical to the `extractTrans` method except it takes an already-parsed node and not a string. Moved out of the class.

src/lexers/jsx-lexer.js
```diff
@@ -1,120 +1,69 @@
...
-      '<Interpolate([^>]*\\s' + this.attr + '[^>]*)\\/?>',
-      'gi'
-    )
-
-    while (matches = regex.exec(content)) {
-      const attrs = this.parseAttributes(matches[1])
-      const key = attrs.keys
-      if (key) {
-        this.keys.push({ ...attrs.options, key })
+export const JSXParserExtension = Object.assign({}, walk.base, {
```
MynockSpit: This is fun. There's no built-ins for walking the JSX part of the tree. This defines what to do when you hit one of these new elements. We're taking it and walk.base and making a new object to use as the traps.

src/lexers/javascript-lexer.js
```diff
@@ -57,8 +58,45 @@ export default class JavascriptLexer extends BaseLexer {
...
+              }
+            })
+        
+            if (entry.key)
+              that.keys.push(entry)
+          }
         }
-      }
+      },
+      JSXParserExtension
```
MynockSpit: We use 'JSXParserExtension' (see below) as a 'base' according to the acorn docs. It's a thing that defines more traps, and how to respond to them.

karellm:
That was quick, thanks a lot! I will merge this and finish the PR

MynockSpit:
Don't believe it, I worked on it yesterday. 😛

Issue #95, opened 2018-05-16
Exception when trying to parse Trans element JSX with interpolation


doncarron:
- Is this a bug or a feature request? bug
- What version are you running (`0.x` or `1.x`)?  1.0.0-beta13
- Which runner (cli, gulp...)? cli

## What I'm trying to do?

Extract translations from a Trans tag that uses interpolation

## How am I doing it?

./node_modules/.bin/i18next 'trans-tmp/components/*.jsx' -c i18n.config.js

Example React code being parsed:

...
                    <DialogHeaderTitle>
                        <Trans>Consumer Information for {{name}}</Trans>
                    </DialogHeaderTitle>
...

Here is my config file:

module.exports = {
    reactNamespace: true,
    locales: ['fr', 'en'],
    output: 'locales'
};

## What was the result?

  i18next Parser
  --------------
  Input:  'trans-tmp/components/*.jsx'
  Output: locales

  [read]  C:\GIT\react-component-final\trans-tmp\components\Component.jsx
  [read]  C:\GIT\react-component-final\trans-tmp\components\ComponentConsumers.jsx
  [read]  C:\GIT\react-component-final\trans-tmp\components\ComponentDependencies.jsx
  [read]  C:\GIT\react-component-final\trans-tmp\components\ComponentList.jsx
  [read]  C:\GIT\react-component-final\trans-tmp\components\Header.jsx
  [read]  C:\GIT\react-component-final\trans-tmp\components\TitleBar.jsx
  [write] C:\GIT\react-component-final\locales\fr\Consumer Information for <1>{{ name.json
  [write] C:\GIT\react-component-final\locales\fr\Consumer Information for <1>{{ name_old.json
  [write] C:\GIT\react-component-final\locales\fr\translation.json
  [write] C:\GIT\react-component-final\locales\fr\translation_old.json
  [write] C:\GIT\react-component-final\locales\en\Consumer Information for <1>{{ name.json
  [write] C:\GIT\react-component-final\locales\en\Consumer Information for <1>{{ name_old.json
  [write] C:\GIT\react-component-final\locales\en\translation.json
  [write] C:\GIT\react-component-final\locales\en\translation_old.json

  Stats:  6 files were parsed
events.js:183
      throw er; // Unhandled 'error' event

Error: ENOENT: no such file or directory, open 'C:\GIT\react-component-final\locales\fr\Consumer Information for <1>{{ name.json'

## What was I expecting?

Proper output translation key: EX:

[...]

PR #97, opened 2018-05-25
a few minor fixes and tweaks (including a fix for #95)

MynockSpit:
1. Regarding splitting keys on namespaces. It doesn't make sense to split a key if the key was pulled from the content -- a complex sentence (or sentences) would make for a bizarre key (related to #95).

2. When we find `{{ foo: bar }}` replace it with `{{ foo }}` to match the behaviour of i18next-react.

3. Strip comments from expressions (i.e. {/* this gets removed */}) to match behaviour of i18next-react.

4. Typo.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -68,7 +68,8 @@ export default class i18nTransform extends Transform {
       let key = entry.key
       const parts = key.split(this.options.namespaceSeparator)
 
-      if (parts.length > 1) {
+      // make sure we're not pulling a 'namespace' out of a default value
+      if (parts.length > 1 && key !== entry.defaultValue) {
         entry.namespace = parts.shift()
       }
       else if (extension === 'jsx' || this.options.reactNamespace) {
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -6,6 +6,9 @@ const JSXParserExtension = Object.assign({}, walk.base, {
   JSXText(node, st, c) {
     // We need this catch, but we don't need the catch to do anything.
   },
+  JSXEmptyExpression(node, st, c) {
+    // We need this catch, but we don't need the catch to do anything.
+  },
   JSXElement(node, st, c) {
     node.openingElement.attributes.forEach(attr => c(attr, st, attr.type))
     node.children.forEach(child => c(child, st, child.type))
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -58,7 +58,7 @@ export default class JsxLexer extends JavascriptLexer {
               entry.defaultValue = defaultValue
 
               if (!entry.key)
-              entry.key = entry.defaultValue
+                entry.key = entry.defaultValue
             }
 
             if (entry.key)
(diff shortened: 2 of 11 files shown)


src/index.js
```diff
@@ -68,7 +68,8 @@ export default class i18nTransform extends Transform {
       let key = entry.key
       const parts = key.split(this.options.namespaceSeparator)
 
-      if (parts.length > 1) {
+      // make sure we're not pulling a 'namespace' out of a default value
+      if (parts.length > 1 && key !== entry.defaultValue) {
```
MynockSpit: Don't split if the key is the same as the default value. If the key is `My favorite paragraph: Here I am. I lied. I'm over here!` you'd get a file named, `My favorite paragraph.json` with a key of `{ "Here I am": { "I lied": "I'm over here!" } }` which seems very very wrong.

src/lexers/jsx-lexer.js
```diff
@@ -6,6 +6,9 @@ const JSXParserExtension = Object.assign({}, walk.base, {
   JSXText(node, st, c) {
     // We need this catch, but we don't need the catch to do anything.
   },
+  JSXEmptyExpression(node, st, c) {
+    // We need this catch, but we don't need the catch to do anything.
+  },
```
MynockSpit: Another one we don't need to do anything except exist.

src/lexers/jsx-lexer.js
```diff
@@ -55,7 +58,7 @@ export default class JsxLexer extends JavascriptLexer {
               entry.defaultValue = defaultValue
 
               if (!entry.key)
-              entry.key = entry.defaultValue
+                entry.key = entry.defaultValue
```
MynockSpit: Welp. Not intended.

src/lexers/jsx-lexer.js
```diff
@@ -112,9 +115,35 @@ export default class JsxLexer extends JavascriptLexer {
         }
       }
       else if (child.type === 'JSXExpressionContainer') {
+        // strip empty expressions
+        if (child.expression.type === 'JSXEmptyExpression')
+          return {
+            type: 'text',
+            content: ''
+          }
```
MynockSpit: This strips comment expressions `{/* comment */}` and empty expressions. `{}`

src/lexers/jsx-lexer.js
```diff
@@ -112,9 +115,35 @@ export default class JsxLexer extends JavascriptLexer {
...
+            content += originalString.slice(start, prop.key.end)
+            start = prop.value.end
+          })
+          content += originalString.slice(start, child.expression.end) + '}'
+
+          return {
+            type: 'js',
+            content
+          }
+        }
```
MynockSpit: This is a really long way of saying, "for each property, ignore the value."

src/lexers/jsx-lexer.js
```diff
@@ -112,9 +115,35 @@ export default class JsxLexer extends JavascriptLexer {
...
+
+        // slice on the expression so that we ignore comments around it
         return {
           type: 'js',
-          content: originalString.slice(child.start, child.end)
+          content: `{${originalString.slice(child.expression.start, child.expression.end)}}`
```
MynockSpit: We're slicing on the `.expression` b/c if we do that it very neatly strips comments out.

src/lexers/jsx-lexer.js
MynockSpit: This needs a tweak. i18next-react ignores whitespace, so we should too

PR #98, opened 2018-05-26
fixed extra spacing in output; added tests for multiple keys in interpolation

MynockSpit:
i18next-react has some interesting behaviour. I suspect we'll be tweaking this for a while. 😕

Notes for this time:
1. `<Trans>before {{ key1, key2 }} after</Trans>` produces a warn, and a default key of `before  after`.
2. Whitespace and values are ignored. These all produce a default key of `<1>{{key1}}</1>`.
    ```jsx
    <Trans>{{ key1: 'value' }}</Trans>
    <Trans>{{key1: 'other value'}}</Trans>
    <Trans>{{key1}}</Trans>
    <Trans>{{ key1 }}</Trans>
    ``` 
3. If included, a `format` property is ignored for the production of the key. (i.e. `<Trans>{{ key1 }}</Trans>` and `<Trans>{{ key1, format }}</Trans>` both produce a key of `<1>{{key1}}</1>`.

Follow their pattern, this will now produce a warn in the first case also.

### DIFF
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -125,18 +125,24 @@ export default class JsxLexer extends JavascriptLexer {
         // strip properties from ObjectExpressions
         // annoying (and who knows how many other exceptions we'll need to write) but necessary
         else if (child.expression.type === 'ObjectExpression') {
-          let content = '{'
-          let start = child.expression.start
+          // i18next-react only accepts two props, any random single prop, and a format prop
+          // for our purposes, format prop is always ignored
 
-          child.expression.properties.forEach(prop => {
-            content += originalString.slice(start, prop.key.end)
-            start = prop.value.end
-          })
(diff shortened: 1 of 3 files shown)


src/lexers/jsx-lexer.js
```diff
@@ -125,18 +125,24 @@ export default class JsxLexer extends JavascriptLexer {
...
+
+          // more than one property throw a warning in i18next-react, but still works as a key
+          if (nonFormatProperties.length > 1) {
+            this.emit('warning', `The passed in object contained more than one variable - the object should look like {{ value, format }} where format is optional.`)
+
+            return {
+              type: 'text',
+              content: ''
+            }
+          }
```
MynockSpit: Produce a warn if you have multiple non-'filter' properties, and strip them. Otherwise strip any spaces that may be between '{{' and the property names.

karellm:
Looks great thanks! 👍

PR #103, opened 2018-06-08
Correct lineEnding option

lmaillet:
While trying to use the _lineEnding_ option, I've got the following error message
```
      if (lineEnding === '\r\n' || lineEnding === 'crlf') {
      ^
ReferenceError: lineEnding is not defined
```

I've just corrected this piece of code to allow using the _lineEnding_ option..


karellm:
@lmaillet Thanks for the contribution! I just published `1.0.0-beta18` with this fix.

lmaillet:
👍

Issue #99, opened 2018-05-30
Merge old translations


nickasd:
- Is this a bug or a feature request? Feature request
- What version are you running (`0.x` or `1.x`)? 1.0.0-beta14
- Which runner (cli, gulp...)? gulp

1) I think it would be nice if the `translation_old.json` file wasn't replaced each time the parser runs, but instead existing old keys were merged with new old keys. This way the developer could keep running the parser without worrying that old translations get lost and sometime in the future she can go through them and delete the ones are not needed anymore. It happened a couple of times to me that after running the parser once I noticed that some keys didn't get parsed correctly because of typos in my source code and after running it a second time all the old translations were gone.

2) I also think that this file should not be created when there are no old translations, so that when the developers runs the parser and no file is created, she knows that there are no changes to be reviewed.

karellm:
I'm ok with a `removeOldKeys` option that would default to `true`. I'm also ok not creating the old files if empty. Another thing I've meant to implement for a while is to store all the old translations into an array.

If you feel like tackling this, a PR is welcome. My focus is on something else at the moment.

nickasd:
Ok, I will do that.

> Another thing I've meant to implement for a while is to store all the old translations into an array.

What do you mean? Would you rather not preserve the nested json structure?

karellm:
I mean:

```json
{
  "oldKey": [
    "old translation",
    "even older translation"
  ]
}
```

nickasd:
I'm now writing a custom merge function to be used instead of the currently used lodash function `extend` when creating the old catalogue. I have noticed that in `mergeHashes` you are checking if the source and target hashes contain arrays. Can you please explain in what situation this can happen? Because if arrays can really appear in the translation hashes, I would have to structure my update a little bit differently.

karellm:
@nickasd I think this will require a deeper rewrite of the codebase to get this working. First the answer to your question is [here](https://www.i18next.com/translation-function/objects-and-arrays).

I believe that this parser should maintain some sort of a hidden catalog with a specific format. It is currently a limitation for storing all old values, it is also a limitation in implementing #55 

Here is what I've had in mind for a while:

```json
{
  nested: {
    key: {
      paths: [
        'somefile.js'
      ],
      values: [
        "old value",
        "older values"
      ]
    }
  }
}
```

This is a pretty big rewrite though. I might take some time on it this week but no guarantee.

PR #104, opened 2018-06-11
prevent deletion of nested values in the old catalog and creation of …

nickasd:
…an empty old catalog

See issue #99. While waiting for a deeper rewrite, I have fixed the deletion of old catalog values.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,8 +1,10 @@
 import _ from 'lodash'
 
-// Turn an entry for the Parser and turn in into a hash,
-// turning the key path 'foo.bar' into an hash {foo: {bar: ""}}
-// The generated hash can be attached to an optional `target`.
+/**
+ * Turn an entry for the Parser and turn in into a hash,
+ * turning the key path 'foo.bar' into an hash {foo: {bar: ""}}
+ * The generated hash can be attached to an optional `target`.
+ */
 function dotPathToHash(entry, target = {}, options = {}) {
   let path = entry.key
   const separator = options.separator || '.'
@@ -31,10 +33,12 @@ function dotPathToHash(entry, target = {}, options = {}) {
   return _.merge(target, result)
 }
 
-// Takes a `source` hash and make sure its value
-// are pasted in the `target` hash, if the target
-// hash has the corresponding key (or if keepRemoved is true).
-// If not, the value is added to an `old` hash.
+/**
+ * Takes a `source` hash and makes sure its value
+ * is pasted in the `target` hash, if the target
+ * hash has the corresponding key (or if `keepRemoved` is true).
+ * If not, the value is added to an `old` hash.
+ */
 function mergeHashes(source, target = {}, old, keepRemoved = false) {
   old = old || {}
   Object.keys(source).forEach(key => {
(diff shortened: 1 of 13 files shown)


test/parser.test.js
```diff
@@ -210,7 +210,7 @@ describe('parser', () => {
     i18nextParser.end(fakeFile)
   })
 
-  it('creates two files per namespace and per locale', (done) => {
+  it('creates one file per namespace and per locale', (done) => {
```
karellm: Maybe I would add a test to check for `_old` catalog creation then. I saw the test below but the catalog already exist and it tests mostly mergeability of the old catalog.

test/helpers/transferValues.test.js
```diff
@@ -0,0 +1,13 @@
+import { assert } from 'chai'
+import { transferValues } from '../../src/helpers'
+
+describe('transferValues helper function', () => {
```
karellm: I would like a little more tests here to better describe what this helper does.

src/transform.js
```diff
@@ -143,11 +142,11 @@ export default class i18nTransform extends Transform {
         newCatalog = populateHash(oldCatalog, newKeys)
 
         // add keys from the current catalog that are no longer used
-        oldCatalog = _.extend(oldCatalog, oldKeys)
+        transferValues(oldKeys, oldCatalog)
 
         // push files back to the stream
         this.pushFile(namespacePath, newCatalog)
-        if (this.options.createOldCatalogs) {
+        if (this.options.createOldCatalogs && Object.keys(oldCatalog).length > 0) {
```
karellm: This project use `Object.keys(oldCatalog).length` or `_.isEmpty`

src/helpers.js
```diff
@@ -105,6 +112,22 @@ function populateHash(source, target = {}) {
...
 }
 
+/**
+ * Merge `source` into `target` by merging nested dictionaries.
+ */
+function transferValues(source, target) {
+  for (let key in source) {
+    const sourceValue = source[key]
+    const targetValue = target[key]
+    if (typeof sourceValue === 'object' && typeof targetValue === 'object' && !Array.isArray(sourceValue)) {
```
karellm: Can you please break this on multiple line to avoid line above 80 char?

```js
if (
  typeof sourceValue === 'object' && 
  typeof targetValue === 'object' && 
  !Array.isArray(sourceValue)
) {
}
```

src/helpers.js
```diff
@@ -49,7 +53,9 @@ function mergeHashes(source, target = {}, old, keepRemoved = false) {
         keepRemoved
       )
       target[key] = nested.new
-      old[key] = nested.old
+      if (Object.keys(nested.old).length > 0) {
```
karellm: This project favours this format: `Object.keys(nested.old).length` or use `_.isEmpty` since it is already loaded.

test/helpers/transferValues.test.js
nickasd: Do you mean that the tests in `transferValues.test.js` are not enough? What would you like to test additionally?

nickasd:
By the way, I forgot to mention that I have slightly modified the comments for the functions in `helpers.js` in order to make them JSDoc compatible. This way, in another file where you use that function, you can hover it and see the description (at least in VSCode).

karellm:
Thanks, made it to `1.0.0-beta19`

Issue #101, opened 2018-05-30
Log added and removed keys


nickasd:
- Is this a bug or a feature request? Feature request
- What version are you running (`0.x` or `1.x`)? 1.0.0-beta14
- Which runner (cli, gulp...)? gulp

It would be nice if the parser would output a list of keys that have been added and removed. This way one can immediately see if new additions were correctly parsed and translate them immediately without first having to compare the output json file with the old version.

Thanks for this great program, it really helps me a lot! I'm happy to make a pull request if desired.

karellm:
Is there a reason you cannot look at the git diff?

nickasd:
I knew this was coming ;-) It's not meant as a replacement for the git diff, rather as a quick way of ensuring that everything went smooth. I guess once you know that everything is set up correctly you will only make use of the git diff, but for new users it would be a nice feedback to know if anything is going on.

Scenario: as I was trying to figure out how this whole thing works, I hadn't set up everything correctly and I had to run the program several times and inspect the output json file every time to check if something happened. Also as I was debugging #100 and editing the Handlebars lexer's source code I had to inspect the output after each change to test if the second key on the same line still hadn't got parsed. Instead it would have been nice to have at least a line telling me how many (new) keys were parsed.

karellm:
If you feel like making a PR I will review it. I would require this to be an option though.

nickasd:
I was working on this today and wondering if in `mergeHashes()` it's really necessary to add removed keys to both the current and old catalogs when the option `keepRemoved` is set. I think that this is not expected behaviour, but the reason why I'm asking is that this kind of messes up the counts of added/merged/removed keys. What do you think?

karellm:
I think it's ok have it only part of the current catalog. The point of the `_old` catalog is really to avoid loosing data.

nickasd:
Ok great. I also have another question. Is there a specific reason why in `populateHash` you only restore an old key if the current value is the empty string? While testing I noticed that if I remove a key with a default value in the html, run the parser, restore the key with the default value in the html, and run the parser again, the key is kept in the old catalog and never restored. I would have expected it to be restored from the old catalog when I restored it in the html.

karellm:
@nickasd I think your assumption is correct. If you want to make a separate PR with test addressing this issue, I will merge it. Thanks!

karellm:
[...]

PR #105, opened 2018-07-07
Fix merge hashes

nickasd:
As discussed in #101, removes keys from the old catalogue that are restored and fixes the restoration of keys from the old catalogue even if the value in the new catalogue is not empty.


src/helpers.js
```diff
@@ -52,71 +53,42 @@ function mergeHashes(source, target = {}, old, keepRemoved = false) {
...
+        const contextMatch = /_([^_]+)?$/.test(singularKey)
+        const rawKey = singularKey.replace(/_([^_]+)?$/, '')
+  
+        mergeTarget = (
+          (contextMatch && target[rawKey] !== undefined) ||
+          (pluralMatch && target[singularKey] !== undefined) ||
+          keepRemoved
+        )
       }
+      (mergeTarget ? target : old)[key] = source[key]
```
karellm: Can you please rewrite this as a `if/else` statement. This project tries to stay away from ternary operator for consistency.

src/helpers.js
```diff
@@ -52,71 +53,42 @@ function mergeHashes(source, target = {}, old, keepRemoved = false) {
...
-        (pluralMatch && target[singularKey] !== undefined)
-      ) {
-        target[key] = source[key]
-      }
-      else if (keepRemoved) {
-        target[key] = source[key]
-        old[key] = source[key]
+      let mergeTarget
+      if (target[key] !== undefined) {
+        mergeTarget = (typeof source[key] === 'string' || Array.isArray(source[key]))
```
karellm: I'd prefer without the wrapping parentheses here

test/helpers/populateHash.test.js
```diff
@@ -1,73 +0,0 @@
...
-  it('replaces `target` empty keys with `source` ones', (done) => {
-    const source = { key1: 'value1' }
-    const target = { key1: '' }
-    const res    = populateHash(source, target)
-
-    assert.deepEqual(res, { key1: 'value1' })
-    done()
-  })
-
-  it('leaves untouched `target` keys that are not empty', (done) => {
```
karellm: Some of these tests should be ported then don't you think?

nickasd: I have gone through them and they looked very similar if not the same as those in `mergeHashes.test.js`. `mergeHashes` and `populateHash` did basically the same thing except that the latter one would merge only if the target was an empty string, which we said is not necessary. Is there something you would add?

src/helpers.js
nickasd: Ok.

src/helpers.js
nickasd: Sure.

nickasd:
Yes, I removed the `old` argument in `mergeHashes` because the parser used to pass `null` and I thought it was confusing. The old catalog is still merged with `transferValues` as it did in the previous PR. What could be done instead is merging the two functions, but it could take some more time.

test/helpers/populateHash.test.js
karellm: One of my comments wasn't published but with the latest commit, I think we are good.

nickasd: I'm not a git expert, but master and origin/master are ancestors of the branch I created. Why do you think it's not rebased?

karellm: Because there are 15 commits and some of them are rewrite some of my own commits that are already in master. I can take care of the rebase.

karellm:
This is published in `1.0.0-beta20`. Thanks for the PR!

PR #106, opened 2018-07-08
Add verbose option

nickasd:
Solves #101. In order to correctly determine the number of merged keys, I had to slightly change the implementation of `dotPathToHash` and keep duplicate keys in `i18nTransform.addEntry`. As a little bonus I added a warning when duplicate keys with different values are detected, which the developer would probably want to fix.
I also took the liberty to replace `forEach` calls with `for` loops in order to be consistent with the rest of the code and because they are more efficient. Feel free to revert them if you prefer.

P.S.: although the diffs are correct, it seems like all the commits since I forked the repo are listed. Do you know what I'm doing wrong?


karellm:
@nickasd Can you please make PRs against the current master? This PR has 15 commits and 0 file changes. I don't know what's wrong but please make sure you use master as a base.

PR #107, opened 2018-07-08
Add verbose option

nickasd:
Solves #101. In order to correctly determine the number of merged keys, I had to slightly change the implementation of dotPathToHash and keep duplicate keys in i18nTransform.addEntry. As a little bonus I added a warning when duplicate keys with different values are detected, which the developer would probably want to fix.
I also took the liberty to replace forEach calls with for loops in order to be consistent with the rest of the code and because they are more efficient. Feel free to revert them if you prefer.

P.S.: although the diffs are correct, it seems like all the commits since I forked the repo are listed. Do you know what I'm doing wrong?


karellm:
This is still not based on master. Here are some steps you can follow:

1. In your fork, checkout master and pull everything from this repo so your fork master is up to date.
2. Checkout a new branch from the fork master
3. Do your work there and create a PR from this new branch

I believe that because you work from `master` in your fork, it is not fast-forward with this project's `master`. You should always work in branches to avoid this.

src/helpers.js
```diff
@@ -64,21 +78,33 @@ function mergeHashes(source, target, keepRemoved = false) {
...
+          (pluralMatch && target[singularKey] !== undefined)
         )
+        if (isKeyMergeable) {
+          pullCount += 1
+        } else {
+          isKeyMergeable = keepRemoved
+          oldCount += 1
+        }
       }
       if (isKeyMergeable) {
```
karellm: I would write this as follow:

```js
else {
  if (target[key] !== undefined) {
    if (
      typeof source[key] === 'string' ||
      Array.isArray(source[key])
    ) {
      target[key] = source[key]
      mergeCount += 1
    } else {
      old[key] = source[key]
      oldCount += 1
    }
  }
  else {
    // support for plural in keys
    const pluralRegex = /_plural(_\d+)?$/;
    const pluralMatch = pluralRegex.test(key)
    const singularKey = key.replace(pluralRegex, '')

    // support for context in keys
    const contextRegex = /_([^_]+)?$/;
    const contextMatch = contextRegex.test(singularKey)
    const rawKey = singularKey.replace(contextRegex, '')

    if (
      (contextMatch && target[rawKey] !== undefined) ||
      (pluralMatch && target[singularKey] !== undefined)
    ) {
      target[key] = source[key]
      pullCount += 1
    } else {
      old[key] = source[key]
      isKeyMergeable = keepRemoved
      oldCount += 1
    }
  }
}
```

src/transform.js
```diff
@@ -97,21 +105,31 @@ export default class i18nTransform extends Transform {
...
         {
           separator: this.options.keySeparator,
           value: this.options.defaultValue
         }
       )
-    })
+      if (duplicate) {
+        uniqueCount -= 1
+        if (conflict) {
+          warn(`Found same keys with different values: ${entry.key}`)
```
karellm: Could you make this an event via `emit`? And I would only call `warn` if `verbose` is true

test/helpers/dotPathToHash.test.js
```diff
@@ -3,55 +3,79 @@ import { dotPathToHash } from '../../src/helpers'
...
+    done()
+  })
+
+  it('detects duplicate keys with the same value', (done) => {
+    const { target, duplicate, conflict } = dotPathToHash(
+      { key: 'one.two.three' },
+      { one: { two: { three: '' } } },
+    )
+    assert.deepEqual(target, { one: { two: { three: '' } } })
+    assert.isTrue(duplicate)
```
karellm: I find `isTrue` and `isFalse` hard to read. Would you mind using `equal(duplicate, true)`? It has a lower cognitive load since it is how we would write/read a test in regular js: `duplicate === true`. I also like reducing the API we use.

package.json
```diff
@@ -27,7 +27,6 @@
     "eol": "^0.9.1",
     "fs-extra": "^6.0.1",
     "gulp-sort": "^2.0.0",
-    "lodash": "~4.17.4",
```
karellm: 🎉

README.md
```diff
@@ -118,6 +118,7 @@ Option                   | Description
 **output**               | Where to write the locale files relative to the base  | `locales`
 **reactNamespace** <sup>3<sup>| For react file, extract the [defaultNamespace](https://react.i18next.com/components/translate-hoc.html)          | `false`
 **sort**                 | Whether or not to sort the catalog                    | `false`
+**verbose**              | The verbostiy level (0 or 1)                          | `0`
```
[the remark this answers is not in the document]
nickasd: It was a boolean in the beginning, but then I thought that in the future there may be other verbosity levels which print out more stuff, like each single key being added or removed, for example. But I can change this if you insist, personally I don't have a need for other verbosity levels currently.

src/helpers.js
nickasd: Ok.

src/transform.js
nickasd: Yes, I could. Don't you think that this should be logged regardless of the verbosity level? After all, this is sort of a programming error, isn't it?

test/helpers/dotPathToHash.test.js
nickasd: Ok.

package.json
nickasd: 👏

src/transform.js
karellm: I don't see this as an error, it is a conflict. If we have an event, people can listen for it when they need too. It is more flexible than a log. The use case would be someone using this with Gulp for instance, they might not want to pollute the log but still catch the conflict.

README.md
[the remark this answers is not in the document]
karellm: I would make it a boolean for now. We can add the verbosity level when it is actually implemented.

nickasd:
I think I was able to rebase and force push everything correctly 😅 Thank you for your patience!

Issue #96, opened 2018-05-21
JsxLexers doesn't support destructing objects


jare-talend:
- Is this a bug or a feature request? bug
- What version are you running (`0.x` or `1.x`)? 1.x
- Which runner (cli, gulp...)? cli

## What I'm trying to do?
I want to parse a file with destructing objects in.

e.g.

src/test.js

```
import React from 'react';

export default function test({ foo, ...rest }) {
	return <div {...rest} />;
}
```

## How am I doing it?

*Replace this section with your config file or cli command*

Here is my config file:

```
{
  output: "locales",
  "lexers": {
    "js": ["JsxLexer"]
  }
}
```

The command I'm running:

```
i18next src/**/*.js -c ./config.json
```

## What was the result?

i18next triggers an error because acorn-jsx doesn't support destructing objects
https://github.com/RReverser/acorn-jsx/issues/75 
```
$ i18next ./src/**/*.js -c ../../config.json

  i18next Parser
  --------------
  Input:  ./src/**/*.js
  Output: locales

  [read]  test.js
{}
node_modules/acorn/dist/acorn.js:2750
  throw err
  ^

SyntaxError: Unexpected token (3:36)
    at Parser.pp$4.raise (node_modules/acorn/dist/acorn.js:2748:13)
    at Parser.pp.unexpected (node_modules/acorn/dist/acorn.js:644:8)
    at Parser.pp$3.parseIdent (node_modules/acorn/dist/acorn.js:2703:10)
    at Parser.pp$3.parsePropertyName (node_modules/acorn/dist/acorn.js:2487:105)
    at Parser.pp$3.parseProperty (node_modules/acorn/dist/acorn.js:2417:8)
    at Parser.pp$3.parseObj (node_modules/acorn/dist/acorn.js:2371:23)
    at Parser.pp$2.parseBindingAtom (node_modules/acorn/dist/acorn.js:1674:19)
    at Parser.pp$2.parseMaybeDefault (node_modules/acorn/dist/acorn.js:1714:23)
    at Parser.pp$2.parseBindingList (node_modules/acorn/dist/acorn.js:1699:25)
    at Parser.pp$1.parseFunctionParams (node_modules/acorn/dist/acorn.js:1223:22)
error Command failed with exit code 1
```
*Replace this with a description of the result you are getting*

## What was I expecting?

parse the file without an error

karellm:
[...]

PR #108, opened 2018-07-09
Adding support for additional acorn plugins

ledfusion:
This pull request enabled i18next to make use of acorn's additional JS plugins:

https://github.com/acornjs/acorn#existing-plugins

All tests pass properly, and new plugins are only loaded if necessary.


src/lexers/javascript-lexer.js
```diff
@@ -14,8 +18,21 @@ export default class JavascriptLexer extends BaseLexer {
   extract(content) {
     const that = this
 
+    var localAcorn = acorn;
+
+    if(this.acornOptions.plugins) {
+      if(this.acornOptions.plugins.stage3)
```
karellm: To follow the project styleguide, please open/close the if statements with `{}`. Also this project doesn't use `;` at the end of statements.

src/lexers/javascript-lexer.js
```diff
@@ -14,8 +18,21 @@ export default class JavascriptLexer extends BaseLexer {
   extract(content) {
     const that = this
 
+    var localAcorn = acorn;
```
karellm: Please use `let` instead of `var`

MynockSpit: Also, why localAcorn? Don't the inject functions take the input object, modify it, and pass it back? In other words, at line 33, is this true: `localAcorn === acorn`?

ledfusion: @MynockSpit I was not keen on modifying the imported acorn object, which is read-only. The stage3 injector `injectAcornStage3` needs the result to be assigned (line 25). That's why a local variable is created. 

In line 33 we have no guarantee that the object reference will always be the same. A sentence like `return Object.assign({}, acorn, ...)` inside the stage3 injector would break the equality.

karellm:
@ledfusion Thanks for the fixes, it is not fast-forward anymore so you'll need to rebase against master and generate the `dist` again.

Also, I won't merge this without tests, do you plan on writing them?

ledfusion:
Hi @karellm, yes I just made the requested changes and now will pull the latest changes on master.

I have also written tests for the related plugins, but as I mentioned in #96 two of the plugins throw an error from acorn:

```
     TypeError: base[type] is not a function
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.skipThrough (node_modules/acorn/dist/walk.js:183:37)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.NewExpression.base.CallExpression (node_modules/acorn/dist/walk.js:370:3)
...
```

I can leave `acorn-stage3` and `acorn-static-class-property-initializer` disabled, or just skip these two tests until the external issue is reported there and solved.

ledfusion:
As mentioned before, the updated pull request has two testcases commented, as they fail to work properly with acorn.

I have investigated the issue locally and I see that when importing the injectors through `require(...)` instead of `import`  babel, then execution works as expected.

I am investigating what may be the difference

ledfusion:
@karellm No problem, I can rebase.

Regarding the issues, I need to make sure of what's the cause before submitting them. So far, it looks like the pluigns inject their logic into the local acorn instance, but then `walk.simple` is not aware of the new node types.

I don't have much time at this point to keep further investigating. Will do my best to at least add the two plugins that work fine.

PR #109, opened 2018-07-10
Adding support for additional acorn plugins

ledfusion:
Follow-up of #108 

This pull request enables i18next-parser to make use of acorn's additional JS plugins acorn-es7 and acorn-acorn-object-rest-spread.

All tests pass properly, and new plugins are only loaded if necessary. Stage3 and static class properties have been backed out by now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,10 +163,8 @@ Note the presence of a `default` which will catch any extension that is not list
         lexer: 'JavascriptLexer',
         acorn: {
           plugins: {
-            stage3: true,
             jsx: true,
             objectRestSpread: true,
-            staticClassPropertyInitializer: true,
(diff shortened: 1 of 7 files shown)


karellm:
Looks great, thanks!

I quick note, next time, you can rebase and force push to the same PR, no need to create a new one.

tstirrat15:
Why were static properties backed out of this? Is there a way to support adding additional plugins that doesn't require creating a pull request each time?

karellm:
@tstirrat15 There is no way to do that without a PR atm but if you make a PR adding this feature I will merge it. As to why it was left out, I think it was bugging back then (I didn't personally tried)

PR #110, opened 2018-07-12
Made the parser more forward compatible

MynockSpit:
- removed acorn-object-rest-spread (b/c acorn supports it natively)
- added stage3 to support dynamic imports
- made the jsx lexer inherit better from javascript lexer
- added some defaults to the config (note that these are my opinions and are arguable)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "acorn": "^5.5.3",
     "acorn-es7": "^0.1.0",
     "acorn-jsx": "^4.1.1",
-    "acorn-object-rest-spread": "^1.1.0",
+    "acorn-stage3": "^0.6.0",
     "broccoli-plugin": "^1.3.0",
     "cheerio": "^1.0.0-rc.2",
     "colors": "~1.2.0-rc0",
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -1,39 +1,57 @@
-import * as acorn from 'acorn-jsx'
-import injectAcornObjectRestSpread from "acorn-object-rest-spread/inject"
+import * as acorn from 'acorn'
+import injectAcornStage3 from "acorn-stage3/inject"
 import injectAcornEs7 from "acorn-es7"
 import * as walk from 'acorn/dist/walk'
 import BaseLexer from './base-lexer'
+
+const WalkerBase = Object.assign({}, walk.base, {
+  Import(node, st, c) {
+    // We need this catch, but we don't need the catch to do anything.
+  }
+})
 
 export default class JavascriptLexer extends BaseLexer {
   constructor(options = {}) {
     super(options)
 
-    this.acornOptions = { sourceType: 'module', ...options.acorn }
+    this.acornOptions = { 
+      sourceType: 'module', 
+      ecmaVersion: 9,
+      ...options.acorn,
+      plugins: {
+        stage3: true,
+        es7: true,
+        ...(options.acorn ? options.acorn.plugins : {})
+      }
+    }
+
     this.functions = options.functions || ['t']
     this.attr = options.attr || 'i18nKey'
+
+    this.acorn = acorn
+    this.WalkerBase = WalkerBase
+
+    if (this.acornOptions.plugins) {
+      if (this.acornOptions.plugins.stage3) {
+        this.acorn = injectAcornStage3(this.acorn)
+      }
+      if (this.acornOptions.plugins.es7) {
+        injectAcornEs7(this.acorn)
+      }
+    }
   }
 
   extract(content) {
     const that = this
 
-    let localAcorn = acorn
-
-    if (this.acornOptions.plugins) {
-      if (this.acornOptions.plugins.es7) {
-        injectAcornEs7(localAcorn)
-      }
-      if (this.acornOptions.plugins.objectRestSpread) {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "acorn": "^5.5.3",
     "acorn-es7": "^0.1.0",
     "acorn-jsx": "^4.1.1",
-    "acorn-object-rest-spread": "^1.1.0",
+    "acorn-stage3": "^0.6.0",
```
MynockSpit: If we set ecmaVersion to 8 or higher, rest is supported natively. 

Added stage3 to support dynamic imports.

src/lexers/javascript-lexer.js
```diff
@@ -1,39 +1,57 @@
...
+import injectAcornStage3 from "acorn-stage3/inject"
 import injectAcornEs7 from "acorn-es7"
 import * as walk from 'acorn/dist/walk'
 import BaseLexer from './base-lexer'
 
+const WalkerBase = Object.assign({}, walk.base, {
+  Import(node, st, c) {
+    // We need this catch, but we don't need the catch to do anything.
+  }
+})
```
MynockSpit: Import is for dynamic imports. Following the pattern in jsx-lexer of putting extensions in a const before the lexer.

src/lexers/javascript-lexer.js
```diff
@@ -1,39 +1,57 @@
...
+    this.acornOptions = { 
+      sourceType: 'module', 
+      ecmaVersion: 9,
+      ...options.acorn,
+      plugins: {
+        stage3: true,
+        es7: true,
+        ...(options.acorn ? options.acorn.plugins : {})
+      }
+    }
```
MynockSpit: These are my opinions -- ecmaVersion 9 supports the most JS, and defaulting stage3 and es7 means that the user has to worry about less config right out the gate. I understand that this may be a contentious change.

src/lexers/javascript-lexer.js
```diff
@@ -1,39 +1,57 @@
...
     walk.simple(
-      localAcorn.parse(content, this.acornOptions),
+      this.acorn.parse(content, this.acornOptions),
       {
         CallExpression(node) {
           that.expressionExtractor.call(that, node)
         }
-      }
+      },
+      this.WalkerBase
```
MynockSpit: These changes are made here so that we can build a walker for JavaScript, then inherit them better into the JSX lexer (instead of duplicating config).

this.acorn is the acorn that's modified for JavaScript (acorn injections)
this.WalkerBase is the acorn that's modified for JavaScript (walk extensions)

src/lexers/jsx-lexer.js
```diff
@@ -79,7 +85,7 @@ export default class JsxLexer extends JavascriptLexer {
           }
         }
       },
-      JSXParserExtension
+      this.WalkerBase
```
MynockSpit: The changes here are to utilize the inheritable properties from the JavascriptLexer. 

- extend WalkerBase with the JSX Extensions
- add the JSX plugin config value to acornConfig
- inject the JSX plugin to acorn itself

test/lexers/javascript-lexer.test.js
```diff
@@ -85,6 +85,13 @@ describe('JavascriptLexer', () => {
     done()
   })
 
+  it('supports async/await', (done) => {
+    const Lexer = new JavascriptLexer({ acorn: { ecmaVersion: 8 } })
+    const content = 'const data = async () => await Promise.resolve()'
+    Lexer.extract(content)
+    done()
+  })
```
MynockSpit: Technically already supported, but I figured we should have a test for it.

test/lexers/javascript-lexer.test.js
```diff
@@ -94,13 +101,23 @@ describe('JavascriptLexer', () => {
     done()
   })
 
-  it('supports the acorn-object-rest-spread plugin', (done) => {
-    const Lexer = new JavascriptLexer({ acorn: { plugins: { objectRestSpread: true } } })
+  it('supports the spread operator in objects plugin', (done) => {
```
MynockSpit: This is the plugin we no longer need 'cause it's supported as long as you set ecmaVersion to 9 or 10.

src/lexers/javascript-lexer.js
karellm: I agree with your decision, I would make sure it is documented though

karellm:
Thanks for the great PR as always. This made it to `1.0.0-beta23`. There is one more change I want to make that's pretty core and then we should be good to get out of beta.

src/lexers/javascript-lexer.js
MynockSpit: Yeah! I was thinking about changing the table-based documentation to snippet based. For example, from this...

Option                   | Description                                           | Default
---------------------- | ----------------------------------------------------- | ---
**contextSeparator**     | Key separator used in your translation keys           | `_`
**createOldCatalogs**    | Save the \_old files                                  | `true`
... | ... | ...

... to this ...

```js
{
  "contextSeparator": "_",
  // Key separator used in your translation keys

  "createOldCatalogs": true,
  // Save the \_old files

  ...
}
```

Thoughts? It lets give them a full config of all defaults as well as letting us define nested config a little easier and without the markdown table restriction of a single line per cell.

PR #111, opened 2018-07-13
allow input globs to be defined in the config file

MynockSpit:
I got tired of specifying the globs in the command line.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -26,16 +26,6 @@ program.on('--help', function() {
 
 program.parse(process.argv)
 
-var args = program.args || []
-
-var globs = args.map(function (s) {
-  s = s.trim()
-  if (s.match(/(^'.*'$|^".*"$)/)) {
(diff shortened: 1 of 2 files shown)


bin/cli.js
```diff
@@ -48,6 +38,33 @@ if (program.config) {
...
+// otherwise, expect to have it specified in the cli
+else {
+  var globs = args.map(function (s) {
+    s = s.trim()
+    if (s.match(/(^'.*'$|^".*"$)/)) {
+      s = s.slice(1, -1)
+    }
+    return s
+  })
+}
```
MynockSpit: This prefers config value over values in the command line. Should it be the other way? Should it merge them?

karellm: I think it is more common for the cli arguments to have precedence over the config file. Would you mind switching the logic?

MynockSpit: Not at all. Switched!

PR #112, opened 2018-07-22
changes readme from using tables to using snippets

MynockSpit:
Here's the change from tables to snippets. Thoughts?

Note that I kept the lexer details separate from the regular config b/c the defaults are complex.

Also noticed while doing this that the output paths are not completely customizable (no matter what you do you get the translations files in a `$NAMESPACE` folder), so I'd like to add or change functionality to support that. More details on line 165.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,109 +100,141 @@ module.exports = i18n
 
 ## Options
 
-Option                   | Description                                           | Default
----------------------- | ----------------------------------------------------- | ---
-**contextSeparator**     | Key separator used in your translation keys           | `_`
-**createOldCatalogs**    | Save the \_old files                                  | `true`
-**defaultNamespace**     | Default namespace used in your i18next config         | `translation`
-**defaultValue**         | Default value to give to empty keys                   | `''`
-**extension** <sup>1<sup>| Extenstion of the catalogs                            | `.json`
-**filename**  <sup>1<sup>| Filename of the catalogs                              | `'$NAMESPACE'`
-**indentation**          | Indentation of the catalog files                      | `2`
-**keepRemoved**          | Keep keys from the catalog that are no longer in code | `false`
-**keySeparator** <sup>2<sup>| Key separator used in your translation keys           | `.`
-**lexers**               | See below for details                                 | `{}`
-**lineEnding**           | Control the line ending. See options at [eol](https://github.com/ryanve/eol) | `auto`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,109 +100,141 @@ module.exports = i18n
...
-**`HandlebarsLexer` options**
+  "namespaceSeparator": ":",
+  // Namespace separator used in your translation keys   
+  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.
 
-Option        | Description                 | Default
-------------- | --------------------------- | -------
-**functions** | Array of functions to match | `['t']`
+  "output": "locales",
+  // Where to write the locale files relative to the base
```
MynockSpit: For reference (and because I didn't realize until I dug into this) files are output following this pattern: `outputs[n]` + `$NAMESPACE` + `filename` + `extension`.

1. Add a `filepath` config option. (e.g.`$NAMESPACE`)
2. Change `filename` to include the namespace. (e.g. `$NAMESPACE/$LOCALE` )
3. Combine all sections into a single option and throw out the others. (e.g. `locales/$NAMESPACE/$LOCALE.json` )
4. Allow defining of full path in `outputs` (like opt 3) and ignore filename and extension if it's specified in outputs. 

Opt 1 and opt 4 are the most backwards compatible ones.  
Opt 1 makes the config more confusing (what's the difference between `outputs` and `filepath`?). 
Opt 3 reduces the number of options and requires explicitness which are generally good things... but what if the user just wants to customize extension OR filename? (Though I suspect this is a rare case.)  
Opt 4 is less appealing than opt 3 b/c bi-modal is a pain in the ass. (It'll probably be fairly easy here, but still. It's the principle of the thing.)
Opt 2 doesn't appeal at all. It seems like the least-good solution: it changes current behaviour AND it doesn't have the benefit of reducing variables or increasing explicitness.

I'd be most happy with Opt 3 or Opt 4. Thoughts?

karellm: Opt 3 is a clear winner for me. We could also support the extension there (.json vs .yml)`.

And I don't care for backward compatibility, we are still in beta. Things can break.

karellm:
Thanks for the PR :)

PR #113, opened 2018-07-22
Remove `extension` and `filename` options in favor all path config in `output`

MynockSpit:
Wow that was a lot of changes. 

Most of them are documentation (or reference) and test changes. The only actual code change happened in transform.js, and it's pretty minimal. Note that we're relying on node's `path.resolve` to turn 'this/that/thou' into window's-style paths if we're on a windows machine.

Every test that used a customized extension, filename or output needed to be tweaked. That accounts for all the changes to the tests that aren't renames.

The documentation and reference changes I made *seem* correct, but I'm not as familiar with the history of the project.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -9,6 +9,8 @@
 - `namespace` was renamed `defaultNamespace`. It defaults to `translation`.
 - `prefix` was deprecated. Use `output`
 - `suffix` was deprecated. Use `output`
+- `filename` was deprecated. Use `output`
+- `extension` was deprecated. Use `output`
 - catalogs are no longer sorted by default. Set `sort` to `true` to enable this.
 
 ## Improvements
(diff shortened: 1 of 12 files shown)


docs/migration.md
```diff
@@ -7,14 +7,14 @@
 - `ignoreVariables` was deprecated. Keys that are not string litterals now emit a warning
 - `writeOld` was renamed `createOldLibraries`. It defaults to `true`.
 - `namespace` was renamed `defaultNamespace`. It defaults to `translation`.
-- `prefix` was deprecated. Use `filename`
-- `suffix` was deprecated. Use `filename`
+- `prefix` was deprecated. Use `output`
+- `suffix` was deprecated. Use `output`
```
karellm: Can you add:

```
- `filename` was deprecated. Use `output`
- `extension` was deprecated. Use `output`
```

MynockSpit: Yup. Added. 🙂

karellm:
Thanks!

PR #118, opened 2018-08-29
Change description to describe the correct thing

tstirrat15:
Looks like some leftover copypasta.


karellm:
Thanks, while at it would you mind matching the description from the cli?  https://github.com/i18next/i18next-parser/blob/master/bin/cli.js#L14

tstirrat15:
Done

PR #120, opened 2018-08-31
Clarify how relative paths work in config

tstirrat15:
This got me when I was trying to use a configuration file in a `tools` folder in my project root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,10 +160,12 @@ module.exports = {
 
   output: 'locales/$LOCALE/$NAMESPACE.json',
   // Supports $LOCALE and $NAMESPACE injection
-  // Where to write the locale files relative to the base
+  // Where to write the locale files relative to the location
+  // of the configuration file
 
   input: undefined,
   // An array of globs that describe where to look for source files
+  // relative to the location of the configuration file
 
   reactNamespace: false,
   // For react file, extract the defaultNamespace - https://react.i18next.com/components/translate-hoc.html


karellm:
I think this is ambiguous, I wouldn't refer to the config file but the "working directory" which might not match. The internal simply does a `path.resolve(output)`.

tstirrat15:
Working directory as in the directory from which the command is run? That wasn't the behavior that I observed:

```bash
yarn run i18next --config tools/i18next-parser.config.js
```
That command failed when the `input` and `output` keys looked like this:
```javascript
{
  input: 'src/**/*.js',
  output: 'src/locales/$LOCALE/$NAMESPACE.json'
}
```
But it was fine when they looked like this:
```javascript
{
  input: '../src/**/*.js',
  output: '../src/locales/$LOCALE/$NAMESPACE.json'
}
```
That's why I was proposing this change. Does `yarn` have some weird behavior around what it considers the current working directory to be?

karellm:
You are totally correct, I forgot I coded it this way. It threw me off because the behaviour is different if you use the `input` as CLI option (then it's relative to `process.cwd()`. If you use the config file, it is indeed relative to the config file, which is a good thing imo.

karellm:
@tstirrat15 Actually I merged your PR but this is not true for the `output` option, only the `input` option right? Can you confirm that you agree with the changes in this commit: https://github.com/i18next/i18next-parser/commit/f2dd18804c2c702b0eb8d50b93b308bee800eeca

Issue #122, opened 2018-09-03
Failing unit tests for object spread operator in v 1.0.0-beta26


phatgolo:
- Is this a bug or a feature request?
Bug
- What version are you running (`0.x` or `1.x`)? 
1.0.0-beta26
- Which runner (cli, gulp...)?
cli

## What I'm trying to do?
Parse object spread operator

## How am I doing it?
By running your unit tests :)

The command I'm running:
```
npm test
```

## What was the result?
1) JavascriptLexer
       supports the spread operator in objects plugin:
     TypeError: Cannot read property 'type' of undefined
      at Object.base.Pattern (node_modules/acorn/dist/walk.js:301:12)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.ObjectPattern (node_modules/acorn/dist/walk.js:323:5)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.Pattern (node_modules/acorn/dist/walk.js:306:7)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.VariableDeclarator (node_modules/acorn/dist/walk.js:281:3)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.VariableDeclaration (node_modules/acorn/dist/walk.js:277:5)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.skipThrough (node_modules/acorn/dist/walk.js:183:37)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.base.Program.base.BlockStatement (node_modules/acorn/dist/walk.js:195:5)
      at c (node_modules/acorn/dist/walk.js:29:15)
      at Object.simple (node_modules/acorn/dist/walk.js:31:5)
      at JavascriptLexer.extract (src/lexers/javascript-lexer.js:47:10)
      at Context.<anonymous> (test/lexers/javascript-lexer.test.js:107:28)

## What was I expecting?
✓ supports the spread operator in objects plugin

## Aditional information
Please correct me if I'm wrong about this, but it seems like you have removed the "acorn-object-rest-spread" dependency, which leads to this error. It was there in version 1.0.0-beta22.

By re-adding this acorn-plugin to the javascript lexer, the test passes again.

karellm:
I can't reproduce this. Please run the following and let me know what you get:

```
rm -rf node_modules
git pull
npm install
npm test
```

phatgolo:
[...]

PR #123, opened 2018-09-04
Fix object rest spread bug in JavaScript lexer

phatgolo:
A solution for the issue: https://github.com/i18next/i18next-parser/issues/122


karellm:
Thanks for the PR. The actual problem was with the dependencies and the lock file. We don't really need to add the plugin option, it's already being taken care of by the other plugins. This should be fixed in `1.0.0-beta27`.

PR #124, opened 2018-09-05
Add arbitrary Acorn plugins via injector configuration

### LINKED ISSUE #121: Add static properties acorn plugin
# Proposal
My project uses static property syntax, which Acorn doesn't parse natively, which means this package is throwing errors. I'd like to add this to the project.

## One approach
It'd be pretty easy to include this in the same way that the other plugins are already included, which would fix my issue right now.

## Another approach
Provide a way to inject plugins through the configuration file. A proposed approach for the `acorn` configuration block:
```javascript
var injectAcornStaticClassPropertyInitializer = require('acorn-static-class-property-initializer/inject');
...
acorn: {
  injectors: [injectAcornStaticClassPropertyInitializer],
  plugins: {
    staticClassPropertyInitializer: true,
  },
}
...
```
This approach would have the benefit of being forward compatible (i.e. there wouldn't be additional work required to add additional plugins in the future) and also allow you to make the acorn plugins `optionalDependencies` of the project.

Thoughts?

tstirrat15:
Fixes #121 

## Motivation
I want to add the static properties acorn plugin, which requires changes to this repo. This approach is a more general approach than adding that plugin directly (in the way that the es7 and stage3 plugins were added), in that it allows for arbitrary plugins to be added to the acorn parser via an `injectors` key in the `acorn` configuration object.

## Discussion points
* This PR wouldn't be a breaking change. We could remove the default injectors, which would reduce the number of dependencies, but then it would be a breaking change. Thoughts?
* It seems like there could be a few more unit tests, but I'm having trouble thinking of what they would be. Any suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,10 +215,7 @@ Note the presence of a `default` which will catch any extension that is not list
       ecmaVersion: 9, // forward compatibility
       // Allows additional acorn plugins via the exported injector functions
       injectors: [],
-      plugins: {
-        es7: true, // some es7 parsing that's not yet in acorn (decorators)
-        stage3: true // load some stage3 configs not yet in a version
-      }
+      plugins: {},
     }
   }],
 
@@ -232,11 +229,8 @@ Note the presence of a `default` which will catch any extension that is not list
     acorn: {
       sourceType: 'module',
       ecmaVersion: 9, // forward compatibility
-      plugins: {
-        es7: true, // some es7 parsing that's not yet in acorn (decorators)
-        stage3: true, // load some stage3 configs not yet in a version
-        jsx: true // always defaults to true in .jsx files
-      }
+      injectors: [],
+      plugins: {},
     }
   }],
 
@@ -253,19 +247,6 @@ Note the presence of a `default` which will catch any extension that is not list
       jsx: 'Preserve',
       target: 'esnext'
     },
-
-    // acorn config (for more information on the acorn options, see here: https://github.com/acornjs/acorn#main-parser)
-    acorn: {
-      sourceType: 'module',
-      ecmaVersion: 9, // forward compatibility
-      // Allows additional acorn plugins via the exported injector functions
-      injectors: [],
-      plugins: {
-        es7: true, // some es7 parsing that's not yet in acorn (decorators)
-        stage3: true, // load some stage3 configs not yet in a version
-        jsx: true // always defaults to true in .jsx files
-      }
-    }
   }]
 ]
 ```
@@ -278,7 +259,7 @@ import injectAcornStaticClassPropertyInitializer from 'acorn-static-class-proper
 
 injectAcornStaticClassPropertyInitializer(acorn);
 ```
-To support these plugins in this project, there is an `injectors` option on the `acorn` configuration key. To add additional plugins to the `JavascriptLexer` oor `JsxLexer`, the configuration would look something like this:
+To support these plugins in this project, there is an `injectors` option on the `acorn` configuration key. To add additional plugins to the `JavascriptLexer` or `JsxLexer`, the configuration would look something like this:
 ```javascript
 import injectAcornStaticClassPropertyInitializer from 'acorn-static-class-property-initializer/inject';
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,8 @@ If you need to change the defaults, you can do it like so:
 #### Javascript
 The Javascript lexer uses [Acorn](https://github.com/acornjs/acorn) to walk through your code and extract references
 translation functions. If your code uses features not supported natively by Acorn, you can enable support through
-`injectors` and `plugins` configuration. This is an example configuration that adds all non-jsx plugins supported by acorn
+`injectors` and `plugins` configuration. Note that you must install these additional dependencies yourself through
+`yarn` or `npm`; they are not included in this package. This is an example configuration that adds all non-jsx plugins supported by acorn
 at the time of writing:
 ```javascript
 const injectAcornStaticClassPropertyInitializer = require('acorn-static-class-property-initializer/inject');
--- a/README.md
+++ b/README.md
@@ -333,8 +333,6 @@ Default configuration:
 }
 ```
 
-#### Plugins and Lexers
-
 ## Events
 
 The transform emits a `reading` event for each file it parses:
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ module.exports = {
 
 The `lexers` option let you configure which Lexer to use for which extension. Here is the default:
 
-Note the presence of a `default` which will catch any extension that is not listed. There are 3 lexers available: `HandlebarsLexer`, `HTMLLexer` and `JavascriptLexer`. Each has configurations of its own. If you need to change the defaults, you can do it like so:
+Note the presence of a `default` which will catch any extension that is not listed. There are 4 lexers available: `HandlebarsLexer`, `HTMLLexer`, `JavascriptLexer`, and `JsxLexer`. Each has configurations of its own. If you need to change the defaults, you can do it like so:
 
 ```js
 [
@@ -213,6 +213,8 @@ Note the presence of a `default` which will catch any extension that is not list
     acorn: {
       sourceType: 'module',
       ecmaVersion: 9, // forward compatibility
+      // Allows additional acorn plugins via the exported injector functions
+      injectors: [],
       plugins: {
         es7: true, // some es7 parsing that's not yet in acorn (decorators)
         stage3: true // load some stage3 configs not yet in a version
@@ -256,6 +258,8 @@ Note the presence of a `default` which will catch any extension that is not list
     acorn: {
       sourceType: 'module',
       ecmaVersion: 9, // forward compatibility
+      // Allows additional acorn plugins via the exported injector functions
+      injectors: [],
       plugins: {
         es7: true, // some es7 parsing that's not yet in acorn (decorators)
         stage3: true, // load some stage3 configs not yet in a version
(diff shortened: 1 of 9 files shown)


src/lexers/javascript-lexer.js
```diff
@@ -10,35 +10,32 @@ const WalkerBase = Object.assign({}, walk.base, {
   }
 })
 
+const defaultInjectors = [injectAcornEs7, injectAcornStage3]
+const defaultPlugins = { stage3: true, es7: true }
```
karellm: I would get rid of this personally and document what could be a good default setup in the README.

README.md
```diff
@@ -234,6 +238,29 @@ Note the presence of a `default` which will catch any extension that is not list
...
 
+#### Acorn Plugins
+[Acorn](https://github.com/acornjs/acorn) supports plugins, which can be enabled by importing their injector functions and then applying them to the acorn instance:
+```javascript
+import acorn from 'acorn';
+import injectAcornStaticClassPropertyInitializer from 'acorn-static-class-property-initializer/inject';
+
+injectAcornStaticClassPropertyInitializer(acorn);
+```
+To support these plugins in this project, there is an `injectors` option on the `acorn` configuration key. To add additional plugins to the `JavascriptLexer` oor `JsxLexer`, the configuration would look something like this:
```
karellm: I think it would gain clarity if we explicitly said that people need to install the dependency. Also I would document how to get to a sensible default setup, maybe list the main 3-4 plugins and what they do. It would make people's life easier.

src/lexers/javascript-lexer.js
```diff
@@ -10,35 +10,32 @@ const WalkerBase = Object.assign({}, walk.base, {
...
-        stage3: true,
-        es7: true,
+        ...defaultPlugins,
         ...(options.acorn ? options.acorn.plugins : {})
       }
     }
 
     this.functions = options.functions || ['t']
     this.attr = options.attr || 'i18nKey'
+    this.injectors = [...defaultInjectors, ...((options.acorn && options.acorn.injectors) || [])]
```
karellm: Could become `this.injectors = options.acorn && options.acorn.injectors || []` if we remove the defaults

src/lexers/javascript-lexer.js
tstirrat15: Sounds good to me.

README.md
tstirrat15: I like this idea.

src/lexers/javascript-lexer.js
tstirrat15: Yep

karellm:
@tstirrat15 You will need to rebase from master since #125 was merged. The doc for TS might need some adjustments too.

tstirrat15:
Yep. I'll also refactor the extension, since I'll be getting rid of the JSX lexer (since it's really a JS lexer with the JSX plugin attached to it).

karellm:
I think the JSX has enough custom code to be worth keeping no? Can you explain your thought process for getting rid of it? Finally, if this happens I would prefer a separate PR.

tstirrat15:
Oh hmm... it totally does. I didn't read that carefully enough. Good call.

tstirrat15:
Wheee. I need to work on my git-fu.

src/lexers/jsx-lexer.js
```diff
@@ -36,7 +35,13 @@ export default class JsxLexer extends JavascriptLexer {
       ...JSXParserExtension
     })
 
-    this.acorn = injectAcornJsx(this.acorn)
+    try {
+      const injectAcornJsx = require('acorn-jsx/inject')
```
tstirrat15: This is my current approach for requiring optional dependencies. There are [packages](https://github.com/jchip/optional-require) that would allow us to do this with a bit less code, though I suppose those would still require a try/catch.

Does it make sense to have this logic in the constructors?

package.json
```diff
@@ -36,7 +32,13 @@
     "vinyl-fs": "^3.0.2",
     "yamljs": "^0.3.0"
   },
-  "optionalDependencies": {},
+  "optionalDependencies": {
```
tstirrat15: I'm not entirely sure that [optionalDependencies](https://docs.npmjs.com/files/package.json#optionaldependencies) makes sense as a place to put these. Apparently it's typically used for packages that may or may not exist on a particular platform (e.g. a filesystem events polyfill for non-Linux platforms), where the software won't break if that package isn't available.

As is, it would install them by default. Is that behavior that we're trying to avoid? If so, should we just leave these out?

package.json
```diff
@@ -52,7 +54,7 @@
     "chai": "^4.1.2",
     "gulp": "^4.0.0",
     "mocha": "^5.0.0",
-    "typescript": "^3.0.3"
```
tstirrat15: I suppose leaving them in devDependencies would be another option.

tstirrat15:
Think `tsx` and `ts` should be separate lexers?

tstirrat15:
@karellm this is ready for the next round of review

README.md
```diff
@@ -187,23 +187,57 @@ module.exports = {
...
+Note the presence of a `default` which will catch any extension that is not listed.
+There are 5 lexers available: `HandlebarsLexer`, `HTMLLexer`, `JavascriptLexer`,
+`JsxLexer`, and `TypescriptLexer`. Each has configurations of its own.
+If you need to change the defaults, you can do it like so:
+
+#### Javascript
+The Javascript lexer uses [Acorn](https://github.com/acornjs/acorn) to walk through your code and extract references
+translation functions. If your code uses features not supported natively by Acorn, you can enable support through
+`injectors` and `plugins` configuration. This is an example configuration that adds all non-jsx plugins supported by acorn
+at the time of writing:
```
karellm: I would still mention that it's the responsability of the user to `npm install` or `yarn add` the dependencies. They are NOT included in the parser package.

README.md
```diff
@@ -251,21 +307,34 @@ Note the presence of a `default` which will catch any extension that is not list
...
+  html: [{
+    lexer: 'HtmlLexer',
+    attr: 'data-i18n' // Attribute for the keys
+    optionAttr: 'data-i18n-options' // Attribute for the options
   }]
-]
+}
 ```
 
+#### Plugins and Lexers
```
karellm: I believe this could be removed now

package.json
karellm: I would add them to the `devDependencies` not yhe `optionalDependencies`. We need them to run tests but that's it.

package.json
karellm: Yes I prefer that option, it IS a dev dependency, we need it to run tests. In "production", it's up to the user to install the plugins he needs.

src/lexers/jsx-lexer.js
```diff
@@ -36,7 +35,13 @@ export default class JsxLexer extends JavascriptLexer {
...
+      this.acorn = injectAcornJsx(this.acorn)
+    } catch (e) {
+      throw new Error(`You must install acorn-jsx to parse jsx files. `
```
karellm: Can you follow this format which is more consistent with the project:

```
throw new Error(
  'You must install acorn-jsx to parse jsx files. ' +
  'Try running "yarn add acorn-jsx" or "npm install acorn-jsx"'
)
```

src/lexers/jsx-lexer.js
karellm: I'm fine with this!

src/lexers/typescript-lexer.js
```diff
@@ -2,30 +2,20 @@ import JsxLexer from './jsx-lexer'
...
-}
-
 export default class TypescriptLexer extends JsxLexer {
   constructor(options = {}) {
     super(options)
     this.tsOptions = options.tsOptions
+    try {
+      this.typescript = require('typescript')
+    } catch (e) {
+      throw new Error(`You must install typescript to parse TypeScript files. `
```
karellm: Same as above

test/lexers/javascript-lexer.test.js
```diff
@@ -109,15 +100,12 @@ describe('JavascriptLexer', () => {
...
   })
-
-  describe('supports the acorn-stage3 plugin', () => {
-
-    it('supports dynamic imports', (done) => {
-      const Lexer = new JavascriptLexer({ acorn: { ecmaVersion: 6, plugins: { stage3: true } } })
-      const content = 'import("path/to/some/file").then(doSomethingWithData)'
-      Lexer.extract(content)
+  describe('supports additional plugins via injector option', () => {
+    it('provided injectors are called with acorn', (done) => {
```
karellm: the name of the test should read as valid english. For instance: it... `injects provided acorn plugins`

test/lexers/javascript-lexer.test.js
```diff
@@ -91,16 +92,6 @@ describe('JavascriptLexer', () => {
     Lexer.extract(content)
     done()
   })
-
-  it('supports the acorn-es7 plugin', (done) => {
```
karellm: I wouldn't remove this test, that's the point of leaving it as a `devDependency`, so we can run a test and check it works. I would move it in the `describe` block above though.

test/lexers/javascript-lexer.test.js
```diff
@@ -109,15 +100,12 @@ describe('JavascriptLexer', () => {
     ])
     done()
   })
-
-  describe('supports the acorn-stage3 plugin', () => {
-
-    it('supports dynamic imports', (done) => {
```
karellm: Same, please don't remove this test, it was added to support some users and we need to know it still works.

test/lexers/typescript-lexer.test.js
```diff
@@ -206,16 +206,6 @@ describe('TypeScript lexer', () => {
     Lexer.extract(content)
     done()
   })
-
-  it('supports the acorn-es7 plugin', (done) => {
```
karellm: same

test/lexers/typescript-lexer.test.js
```diff
@@ -224,15 +214,4 @@ describe('TypeScript lexer', () => {
     ])
     done()
   })
-
-  describe('supports the acorn-stage3 plugin', () => {
-
-    it('supports dynamic imports', (done) => {
```
karellm: same

tstirrat15:
:+1: I'll take some time today to work on these.

tstirrat15:
@karellm have a look now.

karellm:
Thanks, released as part of `1.0.0-beta29`

kmswic:
This would solve my problem (we're using static class property initializers in our codebase), but it's not working for me with 1.0.0-beta29 - I'm getting `Error: Plugin 'staticClassPropertyInitializer' not found` with the sample config included in the docs.
On debugging, it seems that the changes are indeed present in `src/lexers/javascript-lexer.js`, but not in `dist/lexers/javascript-lexer.js` - are you sure the package was published properly?

karellm:
@kmswic You are right, I forgot to push some files, should work in `1.0.0-beta30`. Please let me know if it works for you.

PR #125, opened 2018-09-06
feat: add support for typescript+tsx

connor4312:
Hi 👋 

This PR adds support for parsing TypeScript and tsx files. TypeScript is not automatically included, it's marked as an optional dependency which the consumer should install if they want to be able to parse TypeScript files.


src/lexers/typescript-lexer.js
```diff
@@ -0,0 +1,36 @@
...
+let loadedTs = null
+function loadTypeScript() {
+  if (loadedTs) {
+    return loadedTs;
+  }
+
+  try {
+    loadedTs = require('typescript')
+  } catch (e) {
+    throw new ParsingError(`You must install typescript to parse TypeScript files. Try running "yarn install typescript"`)
```
karellm: I think you meant `yarn add`  no?

test/lexers/typescript-lexer.test.js
```diff
@@ -0,0 +1,23 @@
+import { assert } from 'chai'
+import { testJavaScriptLexer } from './javascript-lexer.test'
+import { testJsxLexer } from './jsx-lexer.test'
```
karellm: I'm really not a fan of adding any unecessary logic to tests. Would you mind simply duplicating the tests. I don't think DRY principles should apply to tests, simplicity and straightforwardness are far more important.

src/lexers/typescript-lexer.js
connor4312: Ah yep. I'm more of an npm user, had yarn here because I noticed that's what you use for this project 😄

test/lexers/typescript-lexer.test.js
connor4312: Sure thing

connor4312:
Comments addressed 🙂 It looks like #124 if merged first will just be a slight addition to the typescript section of the docs and should not require any code changes; if that gets in first, I'll update this PR.

karellm:
LGTM 👍  Thanks for the contribution!

karellm:
Deployed as part of `1.0.0-beta28`

connor4312:
Awesome thank you!

PR #126, opened 2018-09-06
fix: read the program output option correctly

connor4312:
Hello again 🙂 

I noticed the `output` option was never being read, and instead always output to the current working directory. This fixes that.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -36,7 +36,7 @@ if (program.config) {
   }
 }
 
-var output = config.output || program.output
+var output = config.output || program.output || process.cwd()
 
 var args = program.args || []
 var globs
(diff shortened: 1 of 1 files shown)


karellm:
This is intentional. The transform is responsible of outputting the file in the right place, not the `vfs.dest`. I believe everything should be relative to the `process.cwd()` when ran through the CLI.

If I misunderstood you, can you please describe the issue you ran into and offer reproducible steps ?

On a side note, it made me realize that if the `output` is set via a CLI arg instead of the config file, it is not passed to the transform. I fixed it in master and will release a new version when I merge all the PRs.

connor4312:
Ah okay. I noticed that when passing an `-o` option to the CLI, it would seem to be ignored, but it looks like your fix in https://github.com/i18next/i18next-parser/commit/c936417595c3b57546e2f20b7f7180053f8cfdf7 is the right way to deal with this, I'll close the PR. Thanks!

PR #134, opened 2018-10-15
fix(ts): don't parse tsx for plain ts files

connor4312:
TypeScript is slightly incompatible in some cases with tsx, which led to parse errors for some of our files when it read ts as tsx. This checks the extension and only parses to tsx when appropriate.


karellm:
Thanks for the PR!

PR #135, opened 2018-10-16
FIX CLI options in README

patoroco:
There was bad description about `--config` option in CLI.

I copy & paste description from cli itself (https://github.com/i18next/i18next-parser/blob/a9b843504faf31d0573ba83ee113dc1296afb408/bin/cli.js#L14-L16)


karellm:
Thanks for the PR!

PR #142, opened 2018-11-30
Small improvements to the CLI

sebdiem:
I started using the CLI and was annoyed by a few quirks.
The commits should be self-explanatory.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -36,7 +36,7 @@ if (program.config) {
   }
 }
 
-config.output = config.output || program.output
+config.output = program.output || config.output
 
 var args = program.args || []
 var globs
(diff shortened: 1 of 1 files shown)


karellm:
@sebdiem Thanks for the PR. The changes are pretty straight forward indeed but can you please explain your issue and why you made those changes? It's important for me to understand what's going on before merging this in.

sebdiem:
Hmm forget about 4d458de, I misunderstood the documentation and this commit probably does not make any sense.
In fact the documentation states that if you have a "i18next-parser.config.js" file at the root of your project it will be used automatically right ? Not sure that it is really the case though.

karellm:
- `4d458de`: the default is defined in the code. Do you have an exemple of a config that doesn't match the default from the `README`? I'm not keen to add this require since it isn't suppose to change anything. And the CLI won't auto-load the config file, you must be explicit about it. If something is misleading in the documentation, I'd rather address it there than in the CLI code.
- `372aa3c`: 👍 
- `d369b2b` : 👍

sebdiem:
Thanks for your review.
I updated the PR by removing `4d458de`.

karellm:
Thanks, this made it to `1.0.0-beta32`

PR #147, opened 2019-01-14
Feature/use key as default value

tfaieta:
I've been running into a few issues through the development of this feature and I wanted some advice to see how this could best be implemented. 

So far I've added the `useKeysAsDefaultValue` to the options and then passed that into the `dotPathToHash ` function inside the `helpers.js` file. Then I added this ternary `const newValue = entry.defaultValue || options.value || options.useKeysAsDefaultValue ? keyValue : ''
` in order to use the keyValue if it's true. I figured this is very rough but again, looking for help.
 
My thinking is that everything can happen inside that function? 

Also thanks for your time! 👍

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ Thumbs.db
 *.sublime-*
 *.swp
 /sftp-config.json
+/.vscode
 
 # node files
 node_modules/
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ i18next 'app/**/*.{js,hbs}' 'lib/**/*.{js,hbs}' [-oc]
 
 Multiple globbing patterns are supported to specify complex file selections. You can learn how to write globs [here](https://github.com/isaacs/node-glob). Note that glob must be wrapped with single quotes when passed as arguments.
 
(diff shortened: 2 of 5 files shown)


tfaieta:
Changed 
```
const newValue = entry.defaultValue || options.value || options.useKeysAsDefaultValue ? keyValue : ''
```
to 
```
const newValue = options.useKeysAsDefaultValue ? keyValue : entry.defaultValue || entry.defaultValue || options.value || ''
```

and everything seemed to work, I'm going to write expand on the test to see if I missed any sort of edge case.

tfaieta:
Seems like it's good to go for full review now 👍

karellm:
Sorry for the late answer, I updated some of the code style and merged it. It's now part of `1.0.0-beta34`

wanthalf:
I am getting results of this kind:

`"Error": "translation__!NO_KEY_SEPARATOR!__Error"`

Something wrong with my configuration? I use "false" for both separators, as suggested. "translation" is the default namespace.

wanthalf:
Ok, I see it works with the default separator values, at least :-)
(Still, I guess this should be fixed.)

Issue #152, opened 2019-04-28
Interpolating in Trans is incorrect when used without key


vincaslt:
- Is this a bug or a feature request? **Bug**
- What version are you running (`0.x` or `1.x`)? **1.0.0-beta34**
- Which runner (cli, gulp...)? **CLI**

## What I'm trying to do?

Use <Trans> element with interpolated value.

## How am I doing it?

```
<Trans>
  You will start receiving new bookings at <strong>{{ email }}</strong>.
</Trans>
```

Here is my config file:

```
{
  default: [
    {
      lexer: ['TypescriptLexer'],
      tsOptions: {
        jsx: 'Preserve',
        target: 'esnext'
      }
    }
  ],
  input: ['src/**/*.{ts,tsx}'],
  keySeparator: false,
  locales: ['en', 'lt'],
  namespaceSeparator: false,
  output: 'public/locales/$LOCALE/$NAMESPACE.json',
  useKeysAsDefaultValue: true,
  verbose: true
}

```

The command I'm running:

```
i18next --config i18next-parser.config.js
```

## What was the result?

```
"You will start receiving new bookings at <1><0>{{email}}</0></1>.": "You will start receiving new bookings at <1><0>{{email}}</0></1>."
```

## What was I expecting?

```
"You will start receiving new bookings at <1>{{email}}</1>.": "You will start receiving new bookings at <1>{{email}}</1>."
```

Issue #151, opened 2019-04-28
Disabling separators breaks output


vincaslt:
- Is this a bug or a feature request?
Bug

- What version are you running (`0.x` or `1.x`)? 
i18next-parser@1.0.0-beta34

- Which runner (cli, gulp...)?
CLI

## What I'm trying to do?

I want to use keys as values, but without any characters being interpreted as namespace or key separators.

## How am I doing it?

Here is my config file:

```
module.exports = {
  default: [
    {
      lexer: ['TypescriptLexer', 'JsxLexer'],
      tsOptions: {
        jsx: 'Preserve',
        target: 'esnext'
      }
    }
  ],
  input: ['src/**/*.{ts,tsx}'],
  keySeparator: false,
  locales: ['en', 'lt'],
  namespaceSeparator: false,
  output: 'public/locales/$LOCALE/$NAMESPACE.json',
  useKeysAsDefaultValue: true,
  verbose: true
}
```

The command I'm running:

```
i18next --config i18next-parser.config.js
```

## What was the result?

The keys I am getting are all in the form of:

```
{
  "Enter city": "translation__!NO_KEY_SEPARATOR!__Enter city"
}
```

## What was I expecting?

```
{
  "Enter city": "Enter city"
}
```

I think the problem might be here: https://github.com/i18next/i18next-parser/blob/11cb4590ca02e694a8ad717f759f648fa39d704f/src/helpers.js#L14

It ignores whatever is set in the config it seems and just uses dots. Making this change worked for me:

```
newValue = entry.key.substring(entry.key.indexOf(separator) + separator.length, entry.key.length);
```

karellm:
Thanks for the pointer and PR. This was fixed in `1.0.0-beta35`.

PR #153, opened 2019-05-04
Fix parsing of <Trans>

vincaslt:
I've noticed several problems with how `<Trans>` component is being parsed, and fixed them in my PR  for my own use. Maybe this or some parts of it could be merged into the parser if they are relevant for general use? I would be willing to tweak/fix some stuff in order for this to be merged, if needed.

This fixes following:
- #151 Disabling separators would still include them in the output, only development ones. No separators are included after fixes.
- #152 Changes results like `<0>{{text}}</0>` to `{{text}}` for interpolated values. It seems that `react-i18next` stopped doing that at some point, so keys generated from value are never found by `react-i18next`: https://github.com/i18next/react-i18next/blob/9a0722dfb00ee1a9980879843ad58fceef5ac4e2/src/Trans.js#L98
- Added interpolation for string literals, because that's done by `react-i18next` too. So previous result `Value{' '}Another {'Value'}` is now `Value Another Value`
- Fixed how JSX whitespace is handled according to https://reactjs.org/docs/jsx-in-depth.html#children-in-jsx  (same, done by react-i18next, because they just use react's children)

>JSX removes whitespace at the beginning and ending of a line. It also removes blank lines. New lines adjacent to tags are removed; new lines that occur in the middle of string literals are condensed into a single space.

- Some random small things I found like - fixed some leftover `it.only`, or improved usage of test command.


src/lexers/jsx-lexer.js
```diff
@@ -31,6 +31,10 @@ const JSXParserExtension = {
 export default class JsxLexer extends JavascriptLexer {
   constructor(options = {}) {
     super(options)
+    
+    // TODO: this is default in react-i18next, so maybe here too?
```
vincaslt: What do you think?

vincaslt:
Any news?

sourceRS:
thank you for fixing the parsing of my people :p

eddiewang:
Thanks for the fix!

karellm:
@vincaslt Thanks for the PR and good work. I will review this shortly. 

Are you interested in becoming a maintainer for this project? I've basically stopped coding in my life and very rarely find the motivation to dive back into it. This project is still downloaded thousands of time every week so it would be a shame if it slowly died.

vincaslt:
I could help look after the project somewhat. Although I can't dedicate a lot of time to it, as I have my daily job and other stuff to work on. Not sure what being a maintainer entails.

test/parser.test.js
```diff
@@ -724,6 +724,56 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
+    it('parses keeping basic elements if transSupportBasicHtmlNodes is true', (done) => {
```
karellm: The test description is not very clear

src/lexers/jsx-lexer.js
karellm: I'm fine with that, could you just remove the `// TODO` comment?

vincaslt:
done

karellm:
Thank you, I merged it.

PR #156, opened 2019-05-19
enable all tests again

nickasd:
This enables all tests again, some of them failing.


nickasd:
I got a crash when my html files contained keys like `one` and `one.two` both mapped to a string. This fix overwrites any object that might be contained in the existing map, i.e. replacing the subtree of `one` with the new string or the string of `one` with the new subtree.

karellm:
@nickasd This is non fast-forward at this point. I'm also confused as to why you would have both in your file? `i18next` doesn't support that AFAIK.

nickasd:
In my case because of a user error. This kind of error is also logged as a conflict.

nickasd:
Shall I try to resolve the conflicts again?

karellm:
@nickasd If you need this and want to move to `1.0.0` then yes :) I would just invite you to add a better description and title because it doesn't represent the content of the PR anymore. Or you can close this and reopen one.

nickasd:
I opened a new PR #202 .

PR #157, opened 2019-05-22
feature/use-default-separator: use config keySeparator value

liormb:
The separator is being set [here](https://github.com/i18next/i18next-parser/blob/master/src/helpers.js#L11).

It being set to the key coming from the config file (`i18next-parser.config.js`) and default it to a dot (`.`) if not set.

When it needs to set the new value, base on the key, it is using a dot instead Instead of using the separator key that was set before.

## Why
The parser does not use the defined `keySeparator` when `useKeysAsDefaultValue` set to `true`: 

`i18next-parser.config.js`
```
{
  useKeysAsDefaultValue: true,
  keySeparator: '_',
}
```
```
t('example.com')
```
Will parse as:
```
{
  "example": {
    "com": "example.com"
  }
}
```
Instead of:
```
{
  "example.com": "example.com"
}
```

### DIFF
--- a/test/parser.test.js
+++ b/test/parser.test.js
@@ -833,7 +833,7 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
-    it.only('supports useKeysAsDefaultValue', (done) => {
+    it('supports useKeysAsDefaultValue', (done) => {
       let result
       const i18nextParser = new i18nTransform({
         useKeysAsDefaultValue: true,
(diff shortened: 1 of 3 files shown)


test/parser.test.js
```diff
@@ -833,7 +833,7 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
-    it.only('supports useKeysAsDefaultValue', (done) => {
```
liormb: Removing as all specs are suppressed while using the `only` method

liormb:
Added tests

liormb:
bump @karellm

lbenie:
any updates on this?

currently we have our translations completely broken until this is released!

https://github.com/jive/jiveweb/pull/1865

karellm:
Sorry guys for the lack of follow up, I've been very far from coding recently. I've merged #158 because it was the first in the list, I wish I would have seen this before I went on to write my own test :) 

Let me know if the issue persist in `master`. I will also release `1.0.0-beta35` by the end of day.

Issue #166, opened 2019-08-07
Parsing translation keys from comment blocks


blmoore:
- Is this a bug or a feature request? Potential bug
- What version are you running (`0.x` or `1.x`)? 1.x
- Which runner (cli, gulp...)? cli

## What I'm trying to do?

As described in https://github.com/i18next/i18next-parser/issues/83#issuecomment-375907625, I have a dynamically generated translation key and so am hard-coding all possible values the key will hold at runtime in a comment block, hoping they will be picked up by i18next-parser.

## How am I doing it?

```tsx
{
// t("link.firstlink")
// t("link.secondlink")
}
<SomeElement content={t(`link.${variable}`)} />
```

Here is my config file:

```js
const injectAcornStaticClassPropertyInitializer = require("acorn-static-class-property-initializer/inject");
const injectAcornStage3 = require("acorn-stage3/inject");
const injectAcornEs7 = require("acorn-es7");

module.exports = {
  contextSeparator: "_",
  createOldCatalogs: true,
  defaultNamespace: "translation",
  defaultValue: "",
  keepRemoved: false,
  keySeparator: ".",
  lexers: {
    js: ["JavascriptLexer"], 
    jsx: ["JsxLexer"],
    mjs: ["JavascriptLexer"],
    ts: ["TypescriptLexer"],
    tsx: [
      {
        lexer: "TypescriptLexer",
        attr: "i18nKey", // Attribute for the keys
        tsOptions: {
          jsx: "Preserve",
          target: "esnext"
        },
        acorn: {
          injectors: [
            injectAcornStaticClassPropertyInitializer,
            injectAcornStage3,
            injectAcornEs7
          ],
          plugins: {
            staticClassPropertyInitializer: true,
            stage3: true,
            es7: true
          }
        }
      }
    ],
    default: ["TypeScriptLexer"]
  },
  lineEnding: "auto",
  locales: ["en"],
  namespaceSeparator: ":",
  output: "public/locales/$LOCALE/$NAMESPACE.json",
  input: undefined,
  reactNamespace: false,
  useKeysAsDefaultValue: false,
  verbose: true
};

```

The command I'm running:

```
i18next 'src/components/**/*.tsx' --config ./i18n-parser.config.js
```

## What was the result?

No translation keys detected for `link`

## What was I expecting?

Translation keys added from comments:

```json
{
  "link": {
    "firstlink": "",
    "secondlink": ""
  }
}
```

blmoore:
[...]

Issue #155, opened 2019-05-14
Can't parse dynamic imports


externuz:
- Is this a bug or a feature request?
`bug`
- What version are you running (`0.x` or `1.x`)? 
`"i18next-parser": "1.0.0-beta34"`
`"acorn-jsx": "4.x"`
`"acorn-stage3": "0.6.0"`

- Which runner (cli, gulp...)?
`cli`

## What I'm trying to do?
I am trying to parse my `ts` and `tsx` files

## How am I doing it?
`i18next \"./src/**/*.{ts,tsx}\" -o \"./locales\"`

Here is my config file:

```
module.exports = {
  createOldCatalogs: true,
  keepRemoved: true,
  locales: ["en", "ru"],
  verbose: true,
};
```

The command I'm running:

```
i18next \"./src/**/*.{ts,tsx}\" -o \"./locales\"
```

## What was the result?

```
  [read]  /Users/externuz/code/smpio/smmplanner-front/src/api/smp/apiSmpSocialProvidersGet.ts
/Users/externuz/code/smpio/smmplanner-front/node_modules/i18next-parser/dist/parser.js:70
        }} catch (err) {_didIteratorError = true;_iteratorError = err;} finally {try {if (!_iteratorNormalCompletion && _iterator.return) {_iterator.return();}} finally {if (_didIteratorError) {throw _iteratorError;}}}
                                                                                                                                                                                                  ^

SyntaxError: Unexpected token (5:38)
    at Parser.pp$4.raise (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn/dist/acorn.js:2757:13)
    at Parser.pp.unexpected (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn/dist/acorn.js:647:8)
    at Parser.pp$3.parseExprAtom (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn/dist/acorn.js:2196:10)
    at Parser.parseExprAtom (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn-jsx/inject.js:399:24)
    at Parser.pp$3.parseExprSubscripts (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn/dist/acorn.js:2047:19)
    at Parser.pp$3.parseMaybeUnary (/Users/externuz/code/smpio/smmplanner-front/node_modules/acorn/dist/acorn.js:2024:17)
[...]

PR #180, opened 2019-10-11
Use typescript AST

vincaslt:
All tests passing, and that's about it. I haven't tested if it fixed any issues (like #155 or #166 ), or in real-world projects.

This includes the changes from #153 because I don't think it makes sense do this work twice. It should be possible to merge without conflicts were both these PRs got merged.

This PR rewrites lexers to use typescript AST, so typescript is supported implicitly. I think jsx and javascript lexers could be joined somehow to eliminate the distinction, but not sure.

I would say, this is basically a starting point for any effort to improve this tool. As a first step, by dropping acorn parser, because it needed several other dependencies, some of which are incompatible with each other, and causes problems when upgrading.


karellm:
This branch now needs a rebase :)

vincaslt:
Rebased

test/parser.test.js
```diff
@@ -747,7 +747,7 @@ describe('parser', () => {
         first: '',
         second: '',
         third: {
-          first: 'Hello <1>{{name}}</1>, you have {{count}} unread message. <5>Go to messages</5>.',
+          first: 'Hello <strong>{{name}}</strong>, you have {{count}} unread message. <5>Go to messages</5>.',
```
vincaslt: I believe this is a mistake, it shouldn't parse a basic tag here, will try to fix this.

vincaslt: Fixed

karellm:
I just reviewed the PR, it looks really great. Thanks a lot for putting the work to removing `acorn`.

I would like to make you a collaborator of this project. It doesn't really require more of you than what you are already doing. The only thing I'd ask is that we keep working with PRs with peer reviews and that we don't publish directly to `master`. Let me know if that interests you :)

vincaslt:
Yeah, sure

PR #186, opened 2019-10-30
fix(cli): add default values for `--config` and `--output`

iamandrewluca:
It is never too late to be what you might have been.
George Eliot

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -11,8 +11,8 @@ var vfs           = require('vinyl-fs')
 
 program
 .version(pkg.version)
-.option('-c, --config <path>', 'Path to the config file (default: i18next-parser.config.js)')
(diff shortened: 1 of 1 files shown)


vincaslt:
Thanks for the PR. Defaults should indeed exist. There's one problem I see with this kind of straightforward approach of adding them: `i18next-parser.config.js` file itself can provide a default output location: `output: 'locales/$LOCALE/$NAMESPACE.json'`, however after your change it would always be ignored, because not providing any for CLI command, would override anything set in the config.

iamandrewluca:
@vincaslt I get what you mean. Then we can leave the default for `--config` and remove default for `--output` but also remove the note, that a default is used, it' misleading.

vincaslt:
Thanks, merging it.

PR #194, opened 2019-12-17
add vuejs component lexer

newbie78:
vuejs [have ](https://vuejs.org/v2/guide/syntax.html) template part in it's *.vue components.
this PRs add parsing vue component format. tests included ; )

thanx

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "typescript": "^3.6.4",
     "vinyl": "~2.0.1",
     "vinyl-fs": "^3.0.2",
+    "vue-template-compiler": "^2.6.11",
     "yamljs": "^0.3.0"
   },
   "optionalDependencies": {},
--- a/src/lexers/vue-lexer.js
+++ b/src/lexers/vue-lexer.js
@@ -16,7 +16,7 @@ export default class VueLexer extends BaseLexer {
     keys = keys.concat(Lexer.extract(content))
 
     const compiledTemplate = require('vue-template-compiler').compile(content).render
-    const Lexer2 = new JavascriptLexer({functions:this.functions})
+    const Lexer2 = new JavascriptLexer({ functions: this.functions })
     Lexer2.on('warning', warning => this.emit('warning', warning))
     keys = keys.concat(Lexer2.extract(compiledTemplate))
 
--- a/src/lexers/vue-lexer.js
+++ b/src/lexers/vue-lexer.js
@@ -1,0 +1,25 @@
+import BaseLexer from './base-lexer'
+import JavascriptLexer from './javascript-lexer.js'
+
+export default class HandlebarsLexer extends BaseLexer {
+  constructor(options = {}) {
+    super(options)
+
+    this.functions = options.functions || ['$t']
+  }
+
+  extract(content, filename) {
+    let keys = []
+
+    const Lexer = new JavascriptLexer()
+    Lexer.on('warning', warning => this.emit('warning', warning))
+    keys = keys.concat(Lexer.extract(content))
+
+    const compiledTemplate = require('vue-template-compiler').compile(content).render
+    const Lexer2 = new JavascriptLexer({functions:this.functions})
+    Lexer2.on('warning', warning => this.emit('warning', warning))
+    keys = keys.concat(Lexer2.extract(compiledTemplate))
+
+    return keys
+  }
+}
--- a/src/lexers/vue-lexer.js
+++ b/src/lexers/vue-lexer.js
@@ -1,7 +1,7 @@
 import BaseLexer from './base-lexer'
 import JavascriptLexer from './javascript-lexer.js'
 
-export default class HandlebarsLexer extends BaseLexer {
+export default class VueLexer extends BaseLexer {
   constructor(options = {}) {
     super(options)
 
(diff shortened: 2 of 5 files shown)


src/lexers/vue-lexer.js
```diff
@@ -0,0 +1,25 @@
+import BaseLexer from './base-lexer'
+import JavascriptLexer from './javascript-lexer.js'
+
+export default class HandlebarsLexer extends BaseLexer {
```
vincaslt: Probably meant `VueLexer`?

newbie78: absolutely right. i'm fix this.

package.json
```diff
@@ -31,6 +31,7 @@
     "typescript": "^3.6.4",
     "vinyl": "~2.0.1",
     "vinyl-fs": "^3.0.2",
+    "vue-template-compiler": "^2.6.11",
```
karellm: I'm questioning whether we should have this be a `devDependency` and require people who rely on Vue to install it to use the plugin. I'm not keen on adding many dependencies to the project. I especially doubt Vue to be used by most.

wdyt @vincaslt

src/parser.js
```diff
@@ -4,8 +4,11 @@ import HandlebarsLexer from './lexers/handlebars-lexer'
 import HTMLLexer from './lexers/html-lexer'
 import JavascriptLexer from './lexers/javascript-lexer'
 import JsxLexer from './lexers/jsx-lexer'
+import VueLexer from './lexers/vue-lexer'
 
 const lexers = {
+  vue: ['VueLexer'],
```
karellm: Can you follow the somewhat alphabetic order?

test/lexers/vue-lexer.test.js
```diff
@@ -0,0 +1,31 @@
+import { assert } from 'chai'
+import VueLexer from '../../src/lexers/vue-lexer'
+
+describe('VueLexer', () => {
+  it('extracts keys from template & js', (done) => {
```
karellm: I would prefer tests to be atomic and test a single thing as much as possible. This could be 2 tests.

src/lexers/vue-lexer.js
```diff
@@ -0,0 +1,25 @@
...
+
+  extract(content, filename) {
+    let keys = []
+
+    const Lexer = new JavascriptLexer()
+    Lexer.on('warning', warning => this.emit('warning', warning))
+    keys = keys.concat(Lexer.extract(content))
+
+    const compiledTemplate = require('vue-template-compiler').compile(content).render
+    const Lexer2 = new JavascriptLexer({functions:this.functions})
```
karellm: Please follow the project coding style (spaces around `{}` and after `:`)

src/lexers/vue-lexer.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  extract(content, filename) {
+    let keys = []
+
+    const Lexer = new JavascriptLexer()
+    Lexer.on('warning', warning => this.emit('warning', warning))
+    keys = keys.concat(Lexer.extract(content))
+
+    const compiledTemplate = require('vue-template-compiler').compile(content).render
```
karellm: I'm not familiar with the Vue compiler but do we need to run 2 lexers? Couldn't we do it once only?

test/lexers/vue-lexer.test.js
```diff
@@ -0,0 +1,31 @@
...
+    const Lexer = new VueLexer()
+    const content = "<template><p>{{ $t('first') }}</p><template><script>export default { mounted() { this.$i18n.t('second'); } }</script>"
```
karellm: The project code style require each loc to be max 100 char long. 

cc @vincaslt a prettifier would be really nice I believe.

test/lexers/vue-lexer.test.js
newbie78: vue files consist of two parts. template & js. i18next can call from both parts. then test must cover  maximal case. I can add two tests for individual cases but the max case should be considered too.

what do you think?

src/lexers/vue-lexer.js
newbie78: my ide not find lint setting in the project. can you provide settings? or "(spaces around {} and after :) " the only requirements?

src/lexers/vue-lexer.js
newbie78: vue compiler return jsx-like template separately. then I run the result through js-lexer.

package.json
vincaslt: Not sure. At least add it to peer dependencies then?

vincaslt:
I think `dist` build should be generated, no?

test/lexers/vue-lexer.test.js
karellm: I'm fine with that.

src/lexers/vue-lexer.js
karellm: I realize that indeed there is no linter or defined coding style. We will likely introduce one soon.

In the meantine, I invite to look at how the code is currently written and follow that.

package.json
newbie78: vue-template-compiler will definitely be in the project on vuejs
but if you need to parse files outside the project, the user will have to put this package himself.

newbie78:
thanx very mutch! )

Issue #161, opened 2019-06-26
reactNamespace doesn't support `useTranslation` hook


luisrudge:
## What I'm trying to do?

https://github.com/i18next/i18next-parser/blob/46301c7d9cb8e37d832d2259db77b0b01ca365ec/src/transform.js#L250-L258

This code won't evaluate `useTranslation` hooks.

## How am I doing it?

```
const { t } = useTranslation('home');
return <h1>{t(‘welcome’)}</h1>;
```

Here is my config file:

```
module.exports = {
  contextSeparator: '_',
  createOldCatalogs: true,
  defaultNamespace: 't',
  indentation: 2,
  keepRemoved: false,
  keySeparator: false,
  namespaceSeparator: false,
  lexers: {
    js: ['JsxLexer'],
    jsx: ['JsxLexer'],
    default: ['JavascriptLexer']
  },
  lineEnding: 'auto',
  locales: ['en', 'pt-br'],
  output: './src/locales/$LOCALE/$NAMESPACE.json',
  input: undefined,
  reactNamespace: true,
  sort: false,
  useKeysAsDefaultValue: true,
  verbose: false
};
```

The command I'm running:

```
i18next 'src/**/*.js' -c i18next-parser.config.js
```

## What was the result?

`t.json`
```
"welcome”: "t__!NO_KEY_SEPARATOR!__welcome”,
```

## What was I expecting?

`home.json`
```
"welcome”: "home__!NO_KEY_SEPARATOR!__welcome”,
```

bitttttten:
Did you make any progress on this? I am trying to read the source about how it extracts the namespace but I am not having too much luck!

bitttttten:
I have a very crude implementation of this :)

I am using https://www.npmjs.com/package/patch-package and i18next-parser@1.0.0-beta34 for this. Include the following as `patches/i18next-parser+1.0.0-beta34.patch`.

```js
diff --git a/node_modules/i18next-parser/dist/lexers/javascript-lexer.js b/node_modules/i18next-parser/dist/lexers/javascript-lexer.js
index bb00d99..902ff52 100644
--- a/node_modules/i18next-parser/dist/lexers/javascript-lexer.js
+++ b/node_modules/i18next-parser/dist/lexers/javascript-lexer.js
@@ -25,6 +25,8 @@ JavascriptLexer = function (_BaseLexer) {_inherits(JavascriptLexer, _BaseLexer);
 
     _this.acorn = acorn;
[...]

PR #195, opened 2020-01-10
Use default namespace from useTranslation hook if supplied

heivo:


heivo:
This is a fix for issue #161

src/lexers/javascript-lexer.js
```diff
@@ -74,9 +74,17 @@ export default class JavascriptLexer extends BaseLexer {
         }
       }
 
+      if (this.defaultNamespace) {
+        entry.namespace = this.defaultNamespace;
```
karellm: For consistency purposes, could you remove the `;`?

heivo: sure, done

heivo:
I added a feature to evaluate namespace overrides in the t-function options (see docs here: https://www.i18next.com/translation-function/essentials).

hschmied:
@heivo if you could just help me out for a minute, I looked at the commit and see that your code checks if a namespace is given in useTranslation 'string' as well as an array of namespaces, in which case the first one is considered default. (lines 80-82 in your javascript-lexer.js commit)

we are using 1.0.6 of the parser, but for whatever reason, when using a namespace array when initializing 't', the parser is not parsing correctly keys without given namespace (falling back on default).

```
const {t, i18n} = useTranslation('localization', {useSuspense: true});
...
<Text>{t('test_fig')}</Text>
```

... correctly puts it into localization.json ...

```
const {t, i18n} = useTranslation(['localization', 'app', 'common'], {useSuspense: true});
...
<Text>{t('test_fig')}</Text>
```

... does not. instead it writes it into translation.json

---

Btw -- the test included with the commit only checks for useTranslate with namespace-string, but not array.

Thanks & best regards!
-H

heivo:
@hschmied I think the problem is more related to lines 90-92: when an array is supplied as the first argument the namespace cannot be read from `node.arguments[0].text` but from `node.arguments[0].elements[0].text`.

Something like this should work:
```js
    if (
      node.expression.escapedText === "useTranslation" &&
      node.arguments.length
    ) {
      const { text, elements } = node.arguments[0];
      if (text) {
        this.defaultNamespace = text;
      } else if (elements && elements.length) {
        this.defaultNamespace = node.arguments[0].elements[0].text;
      }
    }
```

hschmied:
@heivo jep... that did it and works beautifully! How do you want to proceed to get this change in? I mean you fixed it and should get the credit. :)

karellm:
This should be fixed in `1.0.7`.

PR #202, opened 2020-01-29
Check for conflicting keys with different types

nickasd:
I got a crash when my html files contained keys like one and one.two both mapped to a string. This fix overwrites any object that might be contained in the existing map, i.e. replacing the subtree of one with the new string or the string of one with the new subtree.


karellm:
Thanks! It's part of `1.0.2`

nickasd:
Great, thanks!

PR #207, opened 2020-02-11
Do not set a default value in Trans component

### LINKED ISSUE #206: Default value is not respected for <Trans>
- Is this a bug or a feature request?
Bug
- What version are you running (`0.x` or `1.x`)? 
1.0.2
- Which runner (cli, gulp...)?
cli
## What I'm trying to do?

Trying to extract key from 
`<Trans>A text</Trans>`

## How am I doing it?

`i18next --config i18next-parser.config.js`

Here is my config file:

```
{
  contextSeparator: '_',
  createOldCatalogs: false,
  defaultValue: ''
  keepRemoved: false,
  keySeparator: false,
  lexers: {
    hbs: ['HandlebarsLexer'],

    htm: ['HTMLLexer'],
    html: ['HTMLLexer'],

    mjs: ['JavascriptLexer'],
    js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
    ts: ['JavascriptLexer'],
    jsx: ['JsxLexer'],
    tsx: ['JsxLexer'],

    default: ['JavascriptLexer'],
  },

  lineEnding: 'auto',
  locales: ['en', 'it'],

  namespaceSeparator: '::',
  output: 'public/locales/$LOCALE/$NAMESPACE.json',
  input: [`src/**/*.{ts,tsx}`],
  reactNamespace: true,
  sort: true,
  useKeysAsDefaultValue: false,
  verbose: false,
}
```

The command I'm running:

```
i18next --config i18next-parser.config.js
```

## What was the result?

```
{
  "A text": "A text"
}
```

## What was I expecting?

```
{
  "A text": ""
}
```

ramiel:
When Trans component only contains child and no key attribute,
consider it to be the key but not the default value.
Closes #206

### DIFF
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -66,15 +66,20 @@ export default class JsxLexer extends JavascriptLexer {
       const entry = {}
       entry.key = getKey(tagNode)
 
-      const defaultValue = this.nodeToString.call(this, node, sourceText)
+      if(!entry.key) {
+        entry.key = this.nodeToString.call(this, node, sourceText);
+      } else {
+        const defaultValue = this.nodeToString.call(this, node, sourceText)
 
-      if (defaultValue !== '') {
-        entry.defaultValue = defaultValue
+        if (defaultValue !== '') {
+          entry.defaultValue = defaultValue
 
-        if (!entry.key) {
-          entry.key = entry.defaultValue
+          if (!entry.key) {
+            entry.key = entry.defaultValue
+          }
         }
       }
+
 
       const namespace = getPropValue(tagNode, 'ns')
       if (namespace) {
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -66,8 +66,8 @@ export default class JsxLexer extends JavascriptLexer {
       const entry = {}
       entry.key = getKey(tagNode)
 
-      if(!entry.key) {
-        entry.key = this.nodeToString.call(this, node, sourceText);
+      if (!entry.key) {
+        entry.key = this.nodeToString.call(this, node, sourceText)
       } else {
         const defaultValue = this.nodeToString.call(this, node, sourceText)
 
@@ -79,7 +79,6 @@ export default class JsxLexer extends JavascriptLexer {
           }
         }
       }
-
 
       const namespace = getPropValue(tagNode, 'ns')
       if (namespace) {
--- a/dist/lexers/jsx-lexer.js
+++ b/dist/lexers/jsx-lexer.js
@@ -66,13 +66,17 @@ JsxLexer = function (_JavascriptLexer) {_inherits(JsxLexer, _JavascriptLexer);
         var entry = {};
         entry.key = getKey(tagNode);
 
-        var defaultValue = this.nodeToString.call(this, node, sourceText);
+        if (!entry.key) {
+          entry.key = this.nodeToString.call(this, node, sourceText);
+        } else {
+          var defaultValue = this.nodeToString.call(this, node, sourceText);
 
-        if (defaultValue !== '') {
-          entry.defaultValue = defaultValue;
+          if (defaultValue !== '') {
+            entry.defaultValue = defaultValue;
 
-          if (!entry.key) {
-            entry.key = entry.defaultValue;
+            if (!entry.key) {
+              entry.key = entry.defaultValue;
+            }
           }
         }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7233 @@
+{
+  "name": "i18next-parser",
+  "version": "1.0.6",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/runtime": {
+      "version": "7.9.6",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.9.6.tgz",
+      "integrity": "sha512-64AF1xY3OAkFHqOb9s4jpgk1Mm5vDZ4L3acHvAml+53nO1XbXLuDodsVpO4OIUsmemlUHMxNdYMNJmsvOwLrvQ==",
+      "requires": {
+        "regenerator-runtime": "^0.13.4"
+      }
+    },
+    "@sinonjs/commons": {
+      "version": "1.7.2",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.7.2.tgz",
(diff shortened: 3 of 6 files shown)


vincaslt:
This seems to remove testing default value altogether. I think it would be better to make separate tests with key as default value flag=true and one without (like it is now).

karellm:
I agree with @vincaslt, can you it can still parse a `defaultValue`?

src/lexers/jsx-lexer.js
```diff
@@ -54,16 +54,21 @@ export default class JsxLexer extends JavascriptLexer {
       const entry = {}
       entry.key = getKey(tagNode)
 
-      const defaultValue = this.nodeToString.call(this, node, sourceText)
+      if(!entry.key) {
+        entry.key = this.nodeToString.call(this, node, sourceText);
+      } else {
```
vincaslt: It seems that when there is no key, default value is never set.
When `useKeyAsDefaultValue` is true, it should still set default value though

vincaslt:
Looking at the documentation for Trans component in react-i18next, I'm wondering if this change is actually warranted. Doc: [https://react.i18next.com/latest/trans-component](https://react.i18next.com/latest/trans-component)

Example shows Trans value interpreted as a default value:
```
<Trans i18nKey="welcomeUser">
  Hello <strong>{{name}}</strong>.
</Trans>
// JSON -> "welcomeUser": "Hello <1>{{name}}</1>.",
<Trans i18nKey="multiline">
  Some newlines <br/> would be <br/> fine
</Trans>
// JSON -> "multiline": "Some newlines <1/> would be <3/> fine"
```

ramiel:
The Trans value (children), is interpreted as defaultValue only if `i18nKey` attribute is present, otherwise it should be interpreted as the key, which is the aim of this PR

src/lexers/jsx-lexer.js
ramiel: Ok, you're perfectly right. If `useKeyAsDefaultValue` is true, the defaultValue should be set the same as the key. I thought this check was done later  and not in the lexer. How do I access the options? Is in the options passed to the lexer constructor?

ramiel: You never checked the option before my changes. What was the previous behavior? You always set the default value the same as the key, regardless of the `useKeyAsDefaultValue` option, which is what I'm complaining about in the issue. Was it intended?

ramiel: Looking at the code, it seems that the main options, included `useKeyAsDefaultValue`, are not passed to the lexers. The option seems to be used in the transformer, after the file has been parsed by the lexer. So I guess no other test case should be added to the lexers because the fallback of the value to the key is handled elsewhere

ramiel: Any news on this?

karellm: I'm totally fine that you pass `options` as a second argument [here](https://github.com/i18next/i18next-parser/blob/master/src/parser.js#L69). That way it will be easy to access this and do the check in the lexer.

ramiel: Ok, but why? Lexers already works without knowing about those options and the check is done later. So basically your code is working without passing the options to the lexers and I don't see any reason to start doing it now.

vincaslt: Hmm, I agree with your reasoning, it's probably not up to the lexer to decide what default value should be - it's supposed to be responsible only for making sense of the code. It's likely that the flag was never working as intended, not sure. Probably the logic to assign the default from key based on the flag should be in parser then.

vincaslt: However, I'd interpret this in a way that is contrary to yours, if i18nkey is not set, I assume no key is set, and children is always the default value, but in the parser clean up the key to be the default value if missing (not sure how it impacts other parts, and pretty sure that right now key is being set to default value in lexer anyway)

ramiel: I think there's not a single way to view this, it's just a matter of opinions. From [your comment](https://github.com/i18next/i18next-parser/issues/206#issuecomment-589888363) on the issue it seems like also react-i18next shares your opinion. I still prefer my way, after all there's an option in i18next to fallback to the key if the value is missing which would correctly handle this case as well. Feel free to not continue to review this PR and close it and thanks for all of the attention :)

vincaslt: I'm not sure about your exact use case, but I don't think in general that lexer is working incorrectly right now. I assume you want to skip default values altogether when parsing and have them all empty? Maybe just an additional config flag that disables writing default values in parser/transform level is enough?

ramiel: I think there's such an option already, it's `useKeysAsDefaultValue`. It is just not considered when parsing the `<T>` component because the only value present, i.e. `<T>A text</T>` is considered both the key and the value. I'm just saying that "A text" should be considered only as the key, not the value.

> only maybe it shouldn't make assumptions about key from default value

In my opinion it is doing assumptions about the default value from the key.

Again, I really think the problem is how the children of the T component is considered when the attribute `i18nKey` is not present: if the children are considered being the key, the implementation is now wrong. When the children are considered both the default value and the key, the current implementation is right

ramiel: Something else, reading the documentation of react-i18next, it seems to suggest that, when the `i18nKey` is missing, the value (the children) must considered as the key. It doesn't say the value should be considered as the value. You can find this description [here](https://react.i18next.com/latest/trans-component#trans-props)

vincaslt: if key is missing, the children go through a function (defined in config) to become a key, by default I think it's just the same as default value:
```
hashTransKey: function(defaultValue) {
      // return a key based on defaultValue or if you prefer to just remind you should set a key return false and throw an error
    }
```
And I'm fine with key being the same as default value in case it's missing for Trans component, but I don't think the actual default value should then be omitted based on `useKeysAsDefaultValue` flag alone.

ramiel: Ok, it's fine. So let's close this MR in case. I cannot find the `hashTransKey` in config, can you point me to its description?

ramiel: Ok, found it (https://react.i18next.com/legacy-v9/trans-component#additional-options-on-i-18-next-init).Yes, it seems to suggest the value inside the Trans component is the default value

vincaslt: The parser doesn't have this config value though, maybe it makes sense to add it though as it would solve your issue

karellm:
@vincaslt I believe we can close this right?

karellm:
Closing for lack of activity

Issue #213, opened 2020-04-23
Custom lexers


mzvonar:
- Is this a bug or a feature request?
Feature request
- What version are you running (`0.x` or `1.x`)? 
1.0.2
- Which runner (cli, gulp...)?
cli

## What I'm trying to do?

I would like to use custom lexers

## How am I doing it?

I suggest passing a class in config instead of string. I looked into the code and it's not possible to achieve right now.

Here is my proposed config file:

```
const CustomJavascriptLexer = require('./CustomJavascriptLexer');
const CustomJsxLexer = require('./CustomJsxLexer');

module.exports = {
    lexers: {
        js: [CustomJavascriptLexer],
        ts: [CustomJavascriptLexer],
        jsx: [CustomJsxLexer],
        tsx: [CustomJsxLexer],
        default: [CustomJavascriptLexer]
    }
}
```

I can submit a PR if you are ok with this functionality

vincaslt:
Sure

mzvonar:
Hi @vincaslt 
I noticed that when running yarn test it's using glob pattern `test/**/*.test.js` and file `test/parser.test.js` is not run.
Is this by design or is it a typo?

mzvonar:
I stumbled upon this problem before. 
Sadly, glob pattern`test/**/*.test.js`does not match files right in the `test` folder.
For this there need to be two glob patterns declared, eg. `mocha -r babel-register -r babel-polyfill --recursive test/*.test.js test/**/*.test.js`
Should I submit different PR with fix for this?

karellm:
Both the update test command and custom lexers will be in `1.0.3` released within an hour.

MaffooBristol:
This seems like a cool idea but the documentation doesn't really explain much. How is the class meant to be formed? Does it need to be a class or a pure function? What is the input and output? etc etc

karellm:
Check the existing ones: https://github.com/i18next/i18next-parser/tree/master/src/lexers

I don't plan on writing documentation for the internals of the project but you are welcome to do open a PR!

PR #214, opened 2020-04-23
Support passing functions as custom lexers in lexers config

mzvonar:
Added support for custom lexers (#213), that can be defined by passing function instead of string in lexers config.


karellm:
@vincaslt This LGTM, can you do a quest review too?

vincaslt:
Yep, looks good to me too

PR #220, opened 2020-05-19
Fix extraction of default namespace from useTranslation hook with arrays

heivo:
addition to #195

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -102,11 +102,13 @@ export default class JavascriptLexer extends BaseLexer {
       return entry
     }
 
-    if (
-      node.expression.escapedText === 'useTranslation' &&
-      node.arguments.length
-    ) {
-      this.defaultNamespace = node.arguments[0].text
+    if (node.expression.escapedText === "useTranslation" && node.arguments.length) {
+      const { text, elements } = node.arguments[0];
+      if (text) {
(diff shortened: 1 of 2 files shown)


heivo:
CI pipeline fails for node 6 and 8 but I think that's unrelated to my changes.

karellm:
Hey, I had written some code to add `withTranslation` support. It is now part of `1.0.7` and will fix this too. Thanks for the fix, I actually stole your code that was cleaner than mine for the array part ;)

karellm:
Actually I dropped support for node 6 and 8 which are EOL and published `2.0.0`

amitava82:
Is this fixed for JsxLexer? I have jsx inside js files and I'm using JsxLexer. This seems to be not working.

karellm:
@amitava82 I haven't run the tests since the JSX lexers inherit the JS one. I'll look into it or if you want to do it, a PR is welcome!

AlphaByte02:
As it is now this change was not ported to the final release (or was lost with some merge idk).
If u add the test case for this it will fail.
The fix is pretty simple, just add the same code of the "withTranslation" for the "useTranslation".

@karellm any chance that u could look at it, and maybe fix?

Thanks for the work.

karellm:
@AlphaByte02 I have limited bandwidth at the moment, you seem to know what to fix so a PR would be welcome!

PR #222, opened 2020-06-25
Extract translations from comments

vincaslt:
Adds support for extracting translations from comments. Relevant for: #83 #215

#### Checklist

- [x] only relevant code is changed (make a diff before you submit the PR)
- [x] run tests `npm run test`
- [x] tests are included
- [x] documentation is changed or added

PR #223, opened 2020-06-26
Extract translation from comment for both Javascript and Jsx Lexers

tienpham94:
Adds support for extracting translations from comments for both JavascriptLexer and JsxLexer
Fixes https://github.com/i18next/i18next-parser/issues/166

This PR is an extension of https://github.com/i18next/i18next-parser/pull/222 and is to be merged after https://github.com/i18next/i18next-parser/pull/222. In https://github.com/i18next/i18next-parser/pull/222 the JavascriptLexer is extended to add support for extracting translations from comments. This PR extends it and adds the same support for JsxLexer.

I wanted to add comment extraction support to JsxLexer few days ago. Before seeing @vincaslt 's PR, I went with a quite similar approach to support parsing comments in JsxLexer. You can check the approach in this commit https://github.com/i18next/i18next-parser/commit/3337bb2428b480112fdbf344cc6aab0d640be5c4 if you are interested, but my approach only handles single line comment, @vincaslt 's approach is more extensive that it handles also multiline comment, so I switched to using his approach.

```js
// My initial approach
const sourceFile = ts.createSourceFile(
      filename,
      content,
      ts.ScriptTarget.Latest
    )
const comments = ts.getLeadingCommentRanges(
        sourceFile.getText(),
        node.getFullStart()
      )
 if (comments) {
     comments.map((comment) => {
         const commentEntry = sourceFile
            .getFullText()
            .slice(comment.pos, comment.end)

           if (comment.kind === ts.SyntaxKind.SingleLineCommentTrivia) {
            const regExp = / t\(([^)]+)\)/
            var matches = regExp.exec(commentEntry)
            if (matches) {
              const key = matches[1].replace(/['"\\]+/g, '')
              keys.push({ key })
            }
          }
    })
}

```
### Checklist

- [x] tests are included and pass for jsx-lexer.test.js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-# 2.0.0 - latest
+# 2.1.0 - latest
+
+- Support multiline literals #83
+- Parse comments in js #215
+
+# 2.0.0
 
 - Drop support for node 6 and 8 (EOL) #208
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Finally, if you want to make this process even less painful, I invite you to che
 - Creates one catalog file per locale and per namespace
 - Backs up the old keys your code doesn't use anymore in `namespace_old.json` catalog
 - Restores keys from the `_old` file if the one in the translation file is empty
+- Parses comments for static keys to support dynamic key translations.
 - Supports i18next features:
   - **Context**: keys of the form `key_context`
   - **Plural**: keys of the form `key_plural` and `key_0`, `key_1` as described [here](https://www.i18next.com/translation-function/plurals)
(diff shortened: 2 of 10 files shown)


vincaslt:
Good catch, I thought JS lexer would implicitly handle this for Jsx too, but I didn't bother to make sure... I guess ideally it would reuse the same logic, since it seems pretty similar. As an idea - something like a closure in JS lexer which would keep visited comments' state, that Jsx lexer can reuse:

```
  createCommentNodeParser(node) {
    const visitedComments = new Set()
    const keys = []

    return () => {
      s.forEachLeadingCommentRange(
        ...
      )
    }
  }

extract(content, filename = '__default.js') {
    const parseCommentNode = this.createCommentNodeParser.call(this, node)

    const parseTree = (node) => {
      let entry

      parseCommentNode(node)
      ...
    }
  }
```

karellm:
@tienpham94 Thanks for the PR!

@vincaslt I like your suggestion, I'll let you review the code and let me know when it's ready to merge and publish.

tienpham94:
@vincaslt Updated based on your comment!

karellm:
Published as `2.1.1`

arsalansufi:
Nice! I noticed this issue for the first time today as I was adding a comment to a `.tsx` file and it wasn't getting added to my catalog (I was on v2.1.0). Amazing this this fix was rolled out just yesterday. Thanks!

PR #224, opened 2020-06-28
Accept a function for the defaultValue, useKeysAsDefaultValue, and skipDefaultValues configuration options

zach-waggoner:

There are a few scenarios in which it would be helpful to be able to define a function for these configuration options. For example, you may want to generate a default value for the `en` locale, but not for other locales. Also, if you are using natural language keys, `useKeysAsDefaultValue` doesn't do exactly the right thing because it will include the `_{context}` within the default value. Specifying a function for the `defaultValue` makes it possible to remove the context if you wish to do so.

No

### Checklist

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -11,7 +11,7 @@
  * {one: {two: "bla"}}), `"value"` if the same key already exists swith a
  * different value, or `false`.
  */
-function dotPathToHash(locale, entry, target = {}, options = {}) {
+function dotPathToHash(entry, target = {}, options = {}) {
   let path = entry.key
   if (options.suffix || options.suffix === 0) {
     path += `_${options.suffix}`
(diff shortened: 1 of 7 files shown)


src/helpers.js
```diff
@@ -11,24 +11,39 @@
...
   let path = entry.key
   if (options.suffix || options.suffix === 0) {
     path += `_${options.suffix}`
   }
 
   const separator = options.separator || '.'
-  let newValue = entry.defaultValue || options.value || ''
-
-  if (options.skipDefaultValues) {
+  const key = entry.key.substring(
```
karellm, line 21: This feels complicated, I believe we could accomplish the same thing with making only `defaultValue` be a function. Then I'd do something like:

```js
if (typeof options.defaultValue === 'function') {
  // run it
}
else {
  // the current logic
}
```

src/helpers.js
```diff
@@ -11,24 +11,39 @@
  * {one: {two: "bla"}}), `"value"` if the same key already exists swith a
  * different value, or `false`.
  */
-function dotPathToHash(entry, target = {}, options = {}) {
+function dotPathToHash(locale, entry, target = {}, options = {}) {
```
karellm, line 14: I'm not keen with the `locale` argument to be the first one. I would tend to have it included in the options instead.

src/helpers.js
zach-waggoner: I believe there is a use-case for making at least `skipDefaultValues` a function: since `entry.defaultValue` has higher precedence than `options.defaultValue`, there is still no way to skip the default value for entries that have a specified default value for only some locales (unless `skipDefaultValues` can also be a function).

src/helpers.js
zach-waggoner: Agreed.

Maczuga:
Any news on this? Would be great if that was integrated. 

If I get it right, with this I would be able to:
- parse default value as a translation value for example in lang = en? (in my case, the key is always a base language ready text)
- simply put "TODO" or any text as a translation value for any other language?

karellm:
This would need to be rebased then I'm happy to have a look at it again. It seems like it felt into a crack and was forgotten.

karellm:
Merged as #301 and deployed as `3.10.0`

PR #225, opened 2020-06-28
Add transUseDefaultValueFromOptions JsxLexer option

zach-waggoner:

Because `entry.defaultValue` takes precedence over `options.defaultValue`, it currently isn't possible to specify a default value in the options for \<Trans\> components. #207 tried to solve this but it doesn't seem like a consensus was established about the approach. This PR takes a different approach: add a lexer option to JsxLexer that simply turns off setting the default value on the entry. The existing behavior is unchanged when this option is not included.

Yes: #206

### Checklist

### DIFF
--- a/dist/lexers/jsx-lexer.js
+++ b/dist/lexers/jsx-lexer.js
@@ -11,8 +11,10 @@ JsxLexer = function (_JavascriptLexer) {_inherits(JsxLexer, _JavascriptLexer);
     'br',
     'strong',
     'i',
-    'p'];return _this;
+    'p'];
 
+    _this.transUseDefaultValueFromOptions =
+    options.transUseDefaultValueFromOptions || false;return _this;
   }_createClass(JsxLexer, [{ key: 'extract', value: function extract(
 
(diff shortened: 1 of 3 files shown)


karellm:
Thanks for the PR, I'm not keen to add another option to an area that already quite complicated. My vote goes to passing down the `useKeyAsDefaultValue` option to the lexer and update the logic accordingly.

Wdyt @vincaslt

zach-waggoner:
That works too, but I think we'll also need to pass down `skipDefaultValues` and `defaultValue` so those can also be configured for \<Trans\> components. Ultimately I think we would need to reuse the logic of determining the default value that's in `dotPathToHash`.

zach-waggoner:
After thinking about this some more, I’ve realized that I can just use `skipDefaultValues` (which will work since it has higher precedence than `entry.defaultValue`). I also think that I agree with @vincaslt’s point in #207 that the children effectively _is_ the default value, and the `defaultValue` from the options is only a fallback, so I think that the current behavior makes sense.

Feel free to close this PR, or merge it if you think others will find it useful.

PR #229, opened 2020-07-22
fix: Trans element usage with curly braces syntax

sholzmayer:

If the namespace is provided in curly-braces, it will fall back to the default namespace.
```
<Trans ns={'foo'}>bar</Trans>
```

To avoid this, I added a fix for it and as well a test-case for the curly braces syntax.

No


karellm:
Thanks for the PR. It is published as part of `2.1.2`

PR #232, opened 2020-07-27
Correct regexp to extract namespace from withTranslation HOC

### LINKED ISSUE #161: reactNamespace doesn't support `useTranslation` hook

Yuripetusko:

Looks like regexp is outdated using old react-i18next HOC name

No

### Checklist

### DIFF
--- a/test/templating/react.jsx
+++ b/test/templating/react.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { translate, Trans, Interpolate } from 'react-i18next'
+import { withTranslation, Trans, Interpolate } from 'react-i18next'
 
 const bar = () => (
   <div>
@@ -42,4 +42,4 @@ class Test extends React.Component {
   }
 }
 
-export default translate('react')(Test)
+export default withTranslation('react')(Test)
--- a/dist/transform.js
+++ b/dist/transform.js
@@ -299,7 +299,7 @@ i18nTransform = function (_Transform) {_inherits(i18nTransform, _Transform);
 
     content) {
       var reactTranslateRegex = new RegExp(
-      'translate\\((?:\\s*\\[?\\s*)(' + _baseLexer2.default.stringPattern + ')');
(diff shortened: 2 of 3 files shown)


vincaslt:
I think this works as a temporary solution, but I found that namespace parsing is generally broken, I'm attempting to patch it together here: #232

test/templating/react.jsx
```diff
@@ -42,4 +42,4 @@ class Test extends React.Component {
   }
 }
 
-export default translate('react')(Test)
+export default withTranslation('react')(Test)
```
vincaslt, line 45: Same thing should be done with `typescript.tsx` mock file, but you'll also find that the test is invalid :)

karellm:
Closing in favor of #233

PR #233, opened 2020-07-28
Fix namespace parsing

vincaslt:
Addresses the same issue as this PR: #232 (issue #161) only in a different way.

Lexers already kinda extracted namespaces if they found function calls to `withTranslation` or `useTranslation`, but failed to work correctly because the call order mattered. `transform.js` then made a second pass and extracted namespaces using regex, which was outdated and the original PR addressed that.

However, I'd say that lexers should be used to get the namespace. I fix them by doing a second pass on extracted keys to set the namespaces with the one that lexer found during extraction. `transform.js` then only sets the namespaces with default if there was no other namespace identified by lexer.

This would allow to completely remove `reactNamespace` @karellm : https://github.com/i18next/i18next-parser/issues/161#issuecomment-624729169

I didn't date to do it with this PR as it's a breaking change it seems - react `.js` files will stop being extracted without explicit lexer override, but it essentially serves no other purpose now: 
```
lexers: {
  js: ['JsxLexer']
}
```

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -36,6 +36,15 @@ export default class JavascriptLexer extends BaseLexer {
     }
   }
 
+  setNamespaces(keys) {
+    return this.defaultNamespace
+      ? keys.map((entry) => ({
+          ...entry,
+          namespace: entry.namespace || this.defaultNamespace,
+        }))
+      : keys
+  }
+
   extract(content, filename = '__default.js') {
     const keys = []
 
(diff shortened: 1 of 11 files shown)


src/lexers/javascript-lexer.js
```diff
@@ -36,6 +36,15 @@ export default class JavascriptLexer extends BaseLexer {
     }
   }
 
+  setNamespaces(keys) {
+    return this.defaultNamespace
```
karellm, line 40: I'm not a big fan of ternary operators and the code base temps to avoid them. Would you mind use a boring `if() {} else {}` ?

vincaslt:
Done. They were kinda already implicitly tested for withTranslation, but I made it explicit and also tested useTranslation.

jamiechong:
well done! I was literally about to open an issue about this - then discover it has been fixed **3 minutes ago...**

karellm:
Thanks for the great work, it is deployed as `2.2.0`. Would you mind taking a stab at removing `reactNamespace` too? We can then jump to `3.0.0`.

vincaslt:
> Thanks for the great work, it is deployed as `2.2.0`. Would you mind taking a stab at removing `reactNamespace` too? We can then jump to `3.0.0`.

Sure, I'll try to come around it soon.

PR #234, opened 2020-08-02
[WIP] Support const string literals in translation function

### LINKED ISSUE #215: Support variables in translation function
- Is this a bug or a feature request?

Feels like a fairly common use case, so I believe it's a bug.

- What version are you running (`0.x` or `1.x`)? 

1.0.3

- Which runner (cli, gulp...)?

CLI

## What I'm trying to do?

Generate the output for non-literal translation keys in React components.

```
const key = 'some_key'
const t = useTranslation()

return <Comp prop={t(some_key)} />
```

## How am I doing it?

Here is my config file:

```
{
  output: 'translation-keys.yml',
  input: ['src/lib/**/*.jsx'],
}
```

The command I'm running:

```
i18next
```

## What was the result?

There is no translation key with key `some_key`

## What was I expecting?

There is a translation key with key `some_key`

simoneb:
Fixes #215

To introduce support for const string literals in the translation function:

```js
const key = "some_key";
i18next.t(key);
```

https://github.com/i18next/i18next-parser/issues/215

### Checklist

### DIFF
--- a/dist/lexers/javascript-lexer.js
+++ b/dist/lexers/javascript-lexer.js
@@ -34,6 +34,32 @@ JavascriptLexer = function (_BaseLexer) {_inherits(JavascriptLexer, _BaseLexer);
         });
 
       };
+    } }, { key: 'parseVariables', value: function parseVariables(
+
+    node) {
+      var variables = new Map();
+
+      var parseTree = function parseTree(node) {
+        switch (node.kind) {
+          case ts.SyntaxKind.VariableDeclarationList:{
+              if ((ts.NodeFlags.Const & node.flags) !== ts.NodeFlags.Const) break;
+
+              node.declarations.forEach(function (declaration) {
+                if (declaration.initializer.kind == ts.SyntaxKind.StringLiteral) {
+                  variables.set(declaration.name.text, declaration.initializer.text);
+                }
+              });
+            }
+          default:{
+              node.forEachChild(parseTree);
+              break;
+            }}
+
+      };
+
+      node.forEachChild(parseTree);
+
+      return variables;
     } }, { key: 'extract', value: function extract(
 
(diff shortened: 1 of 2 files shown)


karellm:
The direction seems fine to me. Of course I'd need you to write some tests and make sure they pass.

My only concern is that it slows down large code base parsing. It might be worth making it optional. @vincaslt wdyt?

simoneb:
@karellm in its current status it's really just a proof of concept which I put together to familiarize with the typescript compiler API (which has very little documentation, I figured) and prove that this can be done somewhat easily. 

It's currently neither performant nor correct, because it's traversing the tree twice and may lead to false positives because it doesn't account for scoping rules. I suspect that handling the latter properly may lead to fixing the first too, but clearly making this  implementation bulletproof is another level of effort compared to the POC.

I will have to study const variable declaration scoping rules more closely to understand what a proper way to do this would be, so that cases like the variable being out of scope or another variable from an inner scope overriding one from an outer scope can be handled correctly.

karellm:
@simoneb Do you plan to continue working on this?

simoneb:
> @simoneb Do you plan to continue working on this?

**TL;DR** - no, I'm not planning to

**Longer version**

I did more work compared to what's in this PR but I realized that coming up with a correct implementation of this would be:

- considerably more effort than I thought because in order to avoid false positives/negatives it requires writing a full code parser which accounts for function calls, lexical context, etc..
- less valuable than I initially thought because of the added complexity above and the relatively narrow scope this feature would provide, at least when balanced with the complexity

The full work so far is in [this Codesandbox](https://codesandbox.io/s/typescript-compiler-kbvbo) and I'm fairly confident it would be possible to make it work well, but overall I don't think it's worth it for the reasons above and because this library is already providing an implementation which covers only simple cases.

PR #235, opened 2020-08-07
Remove react namespace

vincaslt:
Removed `reactNamespace` and documented alternative way to get the same functionality the flag was providing - interpreting js files as jsx files (e.g. create-react-app projects).

# Checklist

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -36,11 +36,6 @@ export default class Parser extends EventEmitter {
   constructor(options = {}) {
     super(options)
     this.options = options
-
-    if (options.reactNamespace) {
-      lexers.js = lexers.jsx
-    }
-
     this.lexers = { ...lexers, ...options.lexers }
   }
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -36,11 +36,6 @@ export default class Parser extends EventEmitter {
   constructor(options = {}) {
     super(options)
     this.options = options
-
-    if (options.reactNamespace) {
-      lexers.js = lexers.jsx
```
vincaslt, line 41: This line was apparently shady, because it modified the actual `lexers` constant instead of field on an instance. Because of that a test provided a false positive [here](https://github.com/i18next/i18next-parser/pull/235/files#diff-7bf59ee18e32a02823031db5b25b384cR1447), because there was no reactNamespace the test should've failed, but it passed because a test before it overwrote the constant.

PR #236, opened 2020-08-08
Improve Warnings in CLI

austin-schick:

Warnings are currently invisible outside of verbose mode. These are a nice feature, and it would be helpful for users to see these. We also use i18next-parser as part of automated build processes, and having the translation catalog build fail on a warning would be useful in that process.

This PR
- Changes status text in the CLI to blue
- Displays warnings in yellow by default
- Adds a `--fail-on-warnings` flag to exit with a an exit code of 1 on warnings
- Adds a `--suppress-warnings` flag to restore the original behavior

No

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ npm-debug.log
 
 # project files
 .idea/
+
+# manual test files
+test/manual/en
+test/manual/fr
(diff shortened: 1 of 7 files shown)


karellm:
Thanks, I've merged and [tweaked it](https://github.com/i18next/i18next-parser/commit/b5bc9df10ab691e9b20aaf0726f9fc2189cc06a5) a bit. It's not in `3.0.1`.

PR #237, opened 2020-08-11
fix nested mergeHashes

shanehofstetter:

Option keepRemoved seemed not to work properly when set to true, old keys were still deleted.

Don't know, I can create an issue first if needed.

### Checklist


karellm:
I'm a little confused by this PR. There is no failing test case or added test. Can you describe what the issue is a little more?

karellm:
Thanks! This is deployed as `3.1.0`.

Issue #231, opened 2020-07-23
parsing <Trans> in tsx, no value is extracted


Yuripetusko:
##  🐛 Bug Report

Calling  `save-missing` on Trans with new synthax in tsx does not extract values

## To Reproduce 

```
<Trans
                    i18nKey="questionReviews"
                    ns="testimonialsReviews"
                    defaults="Questions or concerns about one of your reviews? Contact us at
                    <a>hello@test.co.uk</a> and we'll
                    be happy to help!"
                    components={{ a: <a href="mailto:hello@test.co.uk">hello@test.co.uk</a>}}></Trans>
```

`"downloadLocizeTranslations": "i18next 'components/**/*.tsx' && locize save-missing --api-key XXX --project-id XXX --format json --path ./public/static/locales"`

```
const i18nConfig = require('./i18n/config');

module.exports = {
  defaultNamespace: i18nConfig.defaultNamespaces[0],
  // Default namespace used in your i18next config

  defaultValue: 'missing value',
  // Default value to give to empty keys

  locales: i18nConfig.supportedLanguages,
  // An array of the locales in your applications

  namespaceSeparator: ':',
  // Namespace separator used in your translation keys
  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.

  output: `${i18nConfig.localesPath}/$LOCALE/$NAMESPACE.json`,
  // Supports $LOCALE and $NAMESPACE injection
  // Supports JSON (.json) and YAML (.yml) file formats
  // Where to write the locale files relative to process.cwd()
  createOldCatalogs: false,
  reactNamespace: false,
};

```

And output I get is:

```
{
  "questionReviews": "missing value",
}

```

## Expected behavior

```js
{
  "questionReviews": "Questions or concerns about one of your reviews? Contact us at <a>hello@test.co.uk</a> and we'll be happy to help!"
}

```

## Your Environment

- *runtime version*: i.e. node v12.16.§
- *i18next version*: i.e. 19.6.1
- *i18next-parser*: "2.1.2",
- *react-i18next*: "^11.7.0",
- *os*: Mac
- *any other relevant information*

vincaslt:
[...]

PR #238, opened 2020-08-11
Parse default value from 'defaults' prop in Trans

vincaslt:
Adds support for `defaults` prop in trans as per https://react.i18next.com/latest/trans-component#alternative-usage-v-11-6-0

Fixes: #231 & #206 

### Checklist


karellm:
Thanks! This is deployed as `3.1.0`.

PR #245, opened 2020-09-30
Add a 'component' option to JSX-lexer

andriystabryn:

Now you have a possibility to set a translation component name (default `Trans`).
User case: 
We've create a custom component for titles where passed the translation namespace and key `<ScreenTitle path="namespace:translationKey" />` as a result the parser always removes translated string from en.json files because it doesn't understand variables. After my changes I can set component to `ScreenTitle` and attr to `path` and parser works well.
The content of ScreenTitle:
```
export const ScreenTitle = ({ path }: Props) => {
    const { t } = useTranslation(getNameSpace(path));
    return <Text style={Style.headerTitleText}>{t(path)}</Text>;
};
```

No

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ Default configuration:
   jsx: [{
     lexer: 'JsxLexer',
     attr: 'i18nKey', // Attribute for the keys
+    component: 'Trans', // Component name
   }],
 }
 ```
--- a/dist/lexers/jsx-lexer.js
+++ b/dist/lexers/jsx-lexer.js
@@ -13,7 +13,8 @@ JsxLexer = function (_JavascriptLexer) {_inherits(JsxLexer, _JavascriptLexer);
     'i',
     'p'];
 
-    _this.omitAttributes = [_this.attr, 'ns', 'defaults'];return _this;
+    _this.omitAttributes = [_this.attr, 'ns', 'defaults'];
+    _this.component = options.component || 'Trans';return _this;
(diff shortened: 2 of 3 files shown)


karellm:
@andriystabryn This is not mergeable, can you please rebase from master?

@vincaslt Do you see any objection to this? I don't use React.

andriystabryn:
@karellm fixed

vincaslt:
> @andriystabryn This is not mergeable, can you please rebase from master?
> 
> @vincaslt Do you see any objection to this? I don't use React.

Looks good to me

karellm:
@andriystabryn I don't see test or documentation though you checked the boxes in the description. Am I missing something?

andriystabryn:
@karellm sorry, I've missed 'added'. I've just check if tests pass, so everything fine, all tests pass.
Do you require a separate test for my case?

karellm:
@andriystabryn Yes I believe we should test changing the option to something else than "Trans" to make sure it works and future changes don't break the functionality.

karellm:
@andriystabryn Can you please add the test so we can merge this? Or if not close the PR?

MortenRickiRasmussen:
This feature would be amazing! ❤️

karellm:
@MortenRickiRasmussen feel free to rebase the PR and to write a test. I don't have the bandwidth to do that atm.

karellm:
@ahoisl I'm closing this PR for lack of activity. Feel free to reopen one with proper testing and rebased from master

Issue #227, opened 2020-07-16
Parser mangles unicode escape sequences


mareksuscak:
## 🐛 Bug Report

This is probably better explained with an example so here you go:

![image](https://user-images.githubusercontent.com/613647/87720932-fca68180-c7b5-11ea-8e01-05b30adb93de.png)

Notice, how `\u00A0` got turned into a mangled sequence of characters.

## To Reproduce

In order to reproduce, just add `\u00A0` to any of your strings in a translation file and run the parser. This is a unicode escape sequence for a non-breaking space character.

## Expected behavior

The unicode escape sequences must stay intact after parser did its job.

## Your Environment

- *runtime version*: node v12
- *i18next version*: 19.6.0
- *os*: Mac

mareksuscak:
For anyone who picks this up - if there's a better way to add a non-breaking space, I'm all ears. Tried a slew of different approaches and most of them didn't work well.

karellm:
This should be fixed as part of `3.4.0`. Thanks @jamesots for the fix.

ricardosouza-logmein:
Version `i18next-parser@9.1.0` presents the same issue

PR #251, opened 2020-12-07
Improve YAML output

jamesots:

When outputting multiline content into a YAML file, it is now output as a multiline
entry, such as this:

```
multi-line-content: |
  <h1>This is multiline content</h1>
  <p>Isn't it nice</p>
```

Instead of like this:
```
multi-line-content: "<h1>This is multiline content\n<p>Isn't it nice?</p>\n"
```

Yes, #227 (although it's a side effect of using a better YAML library, and isn't the issue I was trying to fix!)

### Checklist

- [ x] only relevant code is changed (make a diff before you submit the PR)
- [ x] tests are included and pass (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))


karellm:
That's great thanks! Could you please add a test for the added functionality? And potentially also for the bug in #227 ?

karellm:
Thanks, this is deployed as `3.4.0`

PR #254, opened 2020-12-17
Outputting autoclosed tags when there's no child content

rformato:
Outputting stuff like `<br></br>` is creating issues in react when `transSupportBasicHtmlNodes` is `true`

When parsing autoclosed tags and having `transSupportBasicHtmlNodes = true` the parser outputs the not autoclosed version oh the tag, causing exceptions in react

No 

### Checklist

### DIFF
--- a/test/lexers/jsx-lexer.test.js
+++ b/test/lexers/jsx-lexer.test.js
@@ -317,6 +317,20 @@ describe('JsxLexer', () => {
         assert.equal(Lexer.extract(content)[0].defaultValue, 'Some Content')
         done()
       })
+
+      it('output self-closing when transSupportBasicHtmlNodes is true and element is self-closing', (done) => {
+        const Lexer = new JsxLexer({transSupportBasicHtmlNodes: true})
+        const content = '<Trans>a<br />b</Trans>'
+        assert.equal(Lexer.extract(content)[0].defaultValue, 'a<br />b')
+        done()
+      })
+
+      it('do not output self-closing when transSupportBasicHtmlNodes is true but element is not self-closing', (done) => {
+        const Lexer = new JsxLexer({transSupportBasicHtmlNodes: true})
+        const content = '<Trans>a<strong></strong>b</Trans>'
+        assert.equal(Lexer.extract(content)[0].defaultValue, 'a<strong></strong>b')
+        done()
+      })
     })
   })
 })
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -132,9 +132,9 @@ export default class JsxLexer extends JavascriptLexer {
                 this.transKeepBasicHtmlNodesFor.includes(child.name)
                   ? child.name
                   : index
-              return `<${elementName}>${elemsToString(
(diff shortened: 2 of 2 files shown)


karellm:
If you open a PR, you should add a test that explain that change. Plus your changes are breaking the test suite. Please fix this if you want me to review and merge it.

rformato:
Thanks @karellm, I had left the PR in draft because it was not ready for review, nevertheless you checked it the same, which is great, thanks for your work. I can see the tests failing now, I'll check them asap.

rformato:
@karellm it is now ready for you to review, thanks

test/lexers/jsx-lexer.test.js
```diff
@@ -317,6 +317,20 @@ describe('JsxLexer', () => {
...
+      it('output self-closing when transSupportBasicHtmlNodes is true and element is self-closing', (done) => {
+        const Lexer = new JsxLexer({transSupportBasicHtmlNodes: true})
+        const content = '<Trans>a<br />b</Trans>'
+        assert.equal(Lexer.extract(content)[0].defaultValue, 'a<br />b')
+        done()
+      })
+
+      it('do not output self-closing when transSupportBasicHtmlNodes is true but element is not self-closing', (done) => {
+        const Lexer = new JsxLexer({transSupportBasicHtmlNodes: true})
+        const content = '<Trans>a<strong></strong>b</Trans>'
```
karellm, line 330: I don't see a self-closing tag in `content`. Is that intentional?

karellm (commented):
See my comment. I'm also unsure why would you even want to output `<br></br>` altogether?

test/lexers/jsx-lexer.test.js
rformato: Yes, the tests are to validate that the tag is not rewritten by the parser when there's no child content, both when the developer writes a self-closing or not self-closing tag. Previously a self-closed tag was always outputted as not self-closed, which is the issue I'm trying to solve.

karellm:
I believe you missed my other question. If I understand correctly, you want:

1. That empty tags stay: `<strong></strong>` stays `<strong></strong>`
2. That self-closing tags stay `<br />` stays `<br />`
3. What do you want to happen with `<br></br>`? And why would you even want to write that code in the first place?

rformato:
oh, sorry @karellm, I'll try to explain:

1. correct
2. correct
3. I think the parser should leave that as the developer wrote it. I don't think the parser should "fix" developer errors, but transparently output the tags as they were written. I think it is up to the developer to read error messages triggered later on, at build time or at run time, and fix them. Ideally no developer will write wrong jsx on purpose, but in case it happens, that error will be transparently output to the language files, and more easily understood and fixed. I would prefer by far reading an error message telling the errors I am doing rather that having the parser silently fix them, denying me the opportunity to learn something new and get better.

I hope it makes sense for you. If that does not sound right for you, I'm more than happy to get your advise

karellm:
It is part of `3.6.0`. Thanks for the fix

rformato:
thank you!

PR #261, opened 2021-02-10
Always emit a warning when key is not a string literal

simontreny:

When running `i18next-parser` on my code, I heavily rely on it to find any keys that are built at runtime so I can add comment-hints or rework code to use static keys instead. Unfortunately, `i18next-parser` only emits warnings when the the key is a simple identifier but not when expressions are used.

Examples of code that do not emit the warning right now:
```js
t(`error.${errorCode}`);
t(cond ? "key1" : "key2");
```

This PR fixes that so that warnings will always be emitted.

No

### Checklist

### DIFF
--- a/dist/lexers/javascript-lexer.js
+++ b/dist/lexers/javascript-lexer.js
@@ -128,12 +128,12 @@ JavascriptLexer = function (_BaseLexer) {_inherits(JavascriptLexer, _BaseLexer);
           }
           entry.key = concatenatedString;
         } else {
-          if (keyArgument.kind === ts.SyntaxKind.Identifier) {
-            this.emit(
-            'warning', 'Key is not a string literal: ' +
-            keyArgument.text);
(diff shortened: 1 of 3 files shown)


karellm:
Thanks for the PR. I agree with the problem you are trying to solve. Can you write a test for it though?

simontreny:
> Can you write a test for it though?

Sure, I just added a unit test for it. Thanks a lot for the quick reply! :)

simontreny:
@karellm Gentle ping: wanted to check back with you if this PR looks OK to you 🙂

karellm:
@simontreny I'm on it now! Thanks for the reminder

karellm:
This was just published as `3.7.0`. Thanks for the PR!

simontreny:
Thanks a lot for the merge and the release, that was quick! 🚀

Issue #263, opened 2021-03-01
Allow to disable (warning on) single line


jotoeri:
## 🚀 Feature Proposal

It would be nice, if one could disable a single line from beeing parsed. I could quite well imagine to have special comments for this like e.g. eslint does: `/* i18next-parser-disable-next-line */` or `// i18next-parser-disable-line`

## Motivation
Having custom methods, that put text (i.e. keys) into an array (must be parsed), while some text must be put into the same array but NOT beeing parsed, i would need to disable single lines from beeing parsed.
Further, having the `t()` function then inserting the respective text out of the array into the page as translated text, i have the t-function onto an array-elements, which gives me a warning. For not to have to check each single warning each time (could even be some new ones and i don't want to miss them...?!) it would be cool to disable the parsing/warning on this line after checking once, so i don't have to check it each time again.

## Example
OneFile.js
```
addToArray('This should be translated, but i will store it and use the actual translation later.');

// Next line should not be parsed, as the error will be in english anyways.
addToArray('Fatal Error: {{ error }}', { error });
```
SecondFile.js
```
$arrayFromOneFile.forEach( (element) => {
    // Next line should not be parsed, as i know it will give a warning, but i checked it and the code is right.
    insertIntoPage( i18n.t(element) );
}
```

karellm:
I'm fine with the proposal and I will happily review a PR!

PR #265, opened 2021-03-03
Include Disable-Command

jotoeri:
See #263, i just wanted to give it a go.
Unfortunately i'm quite confused by this typescript implementation of the lexer and the completely different implementations on the different lexers, and i have no idea, if my attempt could be done better. This works now mostly, i just recognized two small problems, which imo shouldn't be too much of a problem in practice, but honestly prevent the feature from beeing presented as error-free:
- If the disable-comment is the very first line of the file, the whole file is skipped. This seems to be a problem of this node-handling, but i have no idea, what exactly happens with those nodes.
- If a code-line with a trailing dot is right before the comment, then the disable-comment is skipped and the line gets parsed even that it should not. (Probably a general problem of the CommentNodeParser?)

Unfortunately i'm about to give up, so if you can tell me some smaller fixes to make this mergable, just tell me. However if this would need some larger rework just close the PR, i just don't have the time to dig deeper into here... 😞 

Partially #263 for Javascript/Jsx/Vue-Lexers

### Checklist

- [?] I ran `yarn build` to compile and prettify the code -- Not yarn, but `npm run build`?! 🤔

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -156,3 +156,12 @@ In recursive mode, it will parse `*.hbs` and `*.js` files and skip `.git` folder
 }))`
 
 This will output your files in the format `$LOCALE/client.$LOCALE.i18n.json` in a `i18n/` directory.
+
+**Ignore a single line from parsing**
+(Currently only on Js/Jsx/Vue-Lexers)
+Using a preceding comment the following line will not get parsed:
+
+```
+/* i18next-parser-disable-next-line */
+i18n.t('This will not get parsed')
+```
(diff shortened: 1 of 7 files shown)


karellm:
@jotoeri Thanks for this but indeed the issues are to big for me to merge this. If you don't plan to work on it I will close it for now. Anyone, feel free to reopen it if you need this.

ravecat:
Hello @karellm! What about status for current functionality? It's useful feature for dynamic keys, when key exists only third-party services

Disable line will allow don't use workaround `//t('key')` before line

karellm:
@ravecat I don't understand your question.

ravecat:
@karellm I mean similar functionality https://i18next-extract.netlify.app/#/comment-hints?id=disable-extraction-on-a-specific-line-or-code-section

PR #296, opened 2021-04-24
Fix #295, cli not running

### LINKED ISSUE #295: New v3.8.0 always fails to run
## 🐛 Bug Report

`i18next-parser` fails to run after upgrading from 3.7.0 to 3.8.0. Downgrading resolves the issue.

Appears to be the `i18next` command ignoring the config. Though if I explicitly give the config file with the `-c` option, it produces the same error.

## To Reproduce

```bash
$ yarn i18n:frontend
yarn run v1.22.10
/workspace/power-configurator/node_modules/i18next-parser/bin/cli.js:38
    throw err
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at Object.resolve (path.js:980:7)
    at Object.<anonymous> (/workspace/power-configurator/node_modules/i18next-parser/bin/cli.js:32:25)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

My `i18next-parser.config.js`:

```js
// i18next-parser.config.js

module.exports = {
  contextSeparator: '_',
  // Key separator used in your translation keys

  createOldCatalogs: true,
  // Save the \_old files

  // Default namespace used in your i18next config

  defaultValue: '',
  // Default value to give to empty keys

  indentation: 2,
  // Indentation of the catalog files

  keepRemoved: false,
  // Keep keys from the catalog that are no longer in code

  keySeparator: '.',
  // Key separator used in your translation keys
  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.

  // see below for more details
  lexers: {
    hbs: ['HandlebarsLexer'],

    htm: ['HTMLLexer'],
    html: ['HTMLLexer'],

    mjs: ['JavascriptLexer'],
    js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
    ts: ['JavascriptLexer'],
    jsx: ['JsxLexer'],
    tsx: ['JsxLexer'],

    default: ['JavascriptLexer']
  },

  lineEnding: 'auto',
  // Control the line ending. See options at https://github.com/ryanve/eol

  locales: ['en', 'fi', 'de', 'fr', 'se', 'ru'],
  // An array of the locales in your applications

  namespaceSeparator: ':',
  // Namespace separator used in your translation keys
  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.

  output: 'public/locales/$LOCALE/$NAMESPACE.json',
  // Supports $LOCALE and $NAMESPACE injection
  // Supports JSON (.json) and YAML (.yml) file formats
  // Where to write the locale files relative to process.cwd()

  input: ['src/**/*.{ts,tsx}'],
  // An array of globs that describe where to look for source files
  // relative to the location of the configuration file

  sort: false,
  // Whether or not to sort the catalog

  skipDefaultValues: false,
  // Whether to ignore default values.

  useKeysAsDefaultValue: false,
  // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
  // This option takes precedence over the `defaultValue` and `skipDefaultValues` options

  verbose: false,
  // Display info about the parsing including some stats

  failOnWarnings: true,
  // Exit with an exit code of 1 on warnings

  customValueTemplate: null
  // If you wish to customize the value output the value as an object, you can set your own format.
  // ${defaultValue} is the default value you set in your translation function.
  // Any other custom property will be automatically extracted.
  //
  // Example:
  // {
  //   message: "${defaultValue}",
  //   description: "${maxLength}", // t('my-key', {maxLength: 150})
  // }
}
```

## Expected behaviour

Generate the string files like v3.7.0 did

## Your Environment

- *runtime version*: v14 LTS
- *i18next version*: 19.9.2
- *os*: Linux (Ubuntu)

MaxLarue:
Cli does not parse cli argument correctly, leading to failure

This makes i18n extraction impossible

Yes #295

### Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -29,18 +29,18 @@ program.parse(process.argv)
 
 var config = {}
 try {
-  config = require(path.resolve(program.config))
+  config = require(path.resolve(program.opts().config))
 } catch (err) {
   if (err.code === 'MODULE_NOT_FOUND') {
-    console.log('  [error] '.red + 'Config file does not exist: ' + program.config)
+    console.log('  [error] '.red + 'Config file does not exist: ' + program.opts().config)
   }
   else {
     throw err
   }
 }
 
-config.output = program.output || config.output || 'locales/$LOCALE/$NAMESPACE.json'
-config.failOnWarnings = program.failOnWarnings || config.failOnWarnings || false
+config.output = program.opts().output || config.output || 'locales/$LOCALE/$NAMESPACE.json'
+config.failOnWarnings = program.opts().failOnWarnings || config.failOnWarnings || false
 
 var args = program.args || []
 var globs
@@ -75,7 +75,7 @@ else if (config.input) {
       negate = '!'
       s = s.substr(1)
     }
-    return negate + path.resolve(path.dirname(path.resolve(program.config)), s)
+    return negate + path.resolve(path.dirname(path.resolve(program.opts().config)), s)
   })
 }
 
@@ -91,7 +91,7 @@ console.log('  i18next Parser'.cyan)
 console.log('  --------------'.cyan)
 console.log('  Input:  '.cyan + args.join(', '))
 console.log('  Output: '.cyan + config.output)
-if (!program.silent) {
+if (!program.opts().silent) {
   console.log()
 }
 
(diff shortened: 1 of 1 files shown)


karellm:
Thanks I just deployed `3.8.1`

PR #301, opened 2021-04-29
Accept a function for the defaultValue, useKeysAsDefaultValue, and skipDefaultValues configuration options

### LINKED ISSUE #299: Conditional `useKeysAsDefaultValue`
## 🚀 Feature Proposal

I would like to enable `useKeysAsDefaultValue` conditionally based on the generated locale.

## Motivation

I use natural keys, so for the source language (English, in my case), I would like to generate files where the key is set to the default value. However, for all other languages, I would like to generate empty values so it's clear which parts still have to be translated.

## Example

```js
skipDefaultValues: true,
useKeysAsDefaultValue: (locale) => locale === defaultLocale ? true : false
```

---

I'd be happy to send a PR for this if it's something you are willing to add :)

queicherius:
This is https://github.com/i18next/i18next-parser/pull/224, rebased on master. I removed some small logic flow changes to keep the code closer to master.

Closes #299 

### Checklist

- [X] tests are included and pass (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [X] I ran `yarn build` to compile and prettify the code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ Thumbs.db
 *.swp
 /sftp-config.json
 /.vscode
+.history
 
 # node files
 node_modules/
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ module.exports = {
 
   defaultValue: '',
   // Default value to give to empty keys
+  // You may also specify a function accepting the locale, namespace, and key as arguments
 
   indentation: 2,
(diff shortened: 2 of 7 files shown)


karellm:
Thanks, deploying as `3.10.0`

Issue #300, opened 2021-04-29
Configurable `pluralSeparator`


queicherius:
## 🚀 Feature Proposal

I would like to be able to configure the plural separator, just like the key, namespace, and context separator.

## Motivation

I use a different plural separator than `_` to prevent potential conflicts with the keys, since I use natural keys that might include that in the text.

## Example

```ts
pluralSeparator: '___',
```

karellm:
You are welcome to make a PR, I'll review and merge it.

PR #302, opened 2021-04-29
Add the ability to configure plural separators

queicherius:

I would like to be able to configure the plural separator, just like the key, namespace, and context separator.

Closes https://github.com/i18next/i18next-parser/issues/300

### Checklist

- [X] tests are included and pass (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [X] I ran `yarn build` to compile and prettify the code

### DIFF
--- a/dist/helpers.js
+++ b/dist/helpers.js
@@ -14,7 +14,7 @@
 function dotPathToHash(entry) {var target = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};var options = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
   var path = entry.keyWithNamespace;
   if (options.suffix || options.suffix === 0) {
-    path += "_".concat(options.suffix);
+    path += "".concat(options.pluralSeparator).concat(options.suffix);
   }
 
   var separator = options.separator || '.';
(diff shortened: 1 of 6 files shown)


dist/helpers.js
```diff
@@ -14,7 +14,7 @@
 function dotPathToHash(entry) {var target = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};var options = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
   var path = entry.keyWithNamespace;
   if (options.suffix || options.suffix === 0) {
-    path += "_".concat(options.suffix);
+    path += "".concat(options.pluralSeparator).concat(options.suffix);
```
karellm, line 17: Why don't you simply do `options.pluralSeparator.concat(options.suffix)` ?

karellm (commented):
One small comment

dist/helpers.js
queicherius: This was was babel generated on build, not sure why it decided on this particular construct 😄

karellm: Oups I didn't even realize I was reviewing babel's generated code, my bad 😄

karellm:
Deployed in `3.11.0`

queicherius:
> Deployed in `3.11.0`

Awesome, thanks!

PR #316, opened 2021-05-11
Update supported node versions

mmorel-35:
Update supported node versions to allow PR #310

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: node_js
 
 node_js:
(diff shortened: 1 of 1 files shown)


karellm:
This it is a breaking change (would require a version bump) and version 10 is still officially maintained. I'm not keen to drop support too fast. Did you look into what exactly the issue is?

mmorel-35:
This is in the changelog of fs-extra, they don't support anymore 10.X

PR #317, opened 2021-05-11
Use Github workflows and add test coverage with codecov.io

mmorel-35:
Use Github workflows instead of travis


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,33 @@
...
+      uses: actions/cache@v2
+      with:
+        path: ~/.npm
+        key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.OS }}-node-
+          ${{ runner.OS }}-
+    - run: npm install
+    - run: npm run build
+    - run: npm run coverage
```
karellm, line 29: Shouldn't it `- run: npm test` too?

karellm (commented):
I added a small comment. Also I believe you need to rebase the PR.

.github/workflows/node.js.yml
mmorel-35: It does, https://github.com/i18next/i18next-parser/pull/317/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R12 
Coverage includes test

mmorel-35:
You shall now be able to add the appropriate badges to node js workflow and one for codecov.io which both require ownership to access.

karellm:
Do you have a pointer to the docs?

mmorel-35:
* https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge
* https://docs.codecov.io/docs/status-badges

PR #332, opened 2021-05-29
Improve behaviour for missing defaults

dhaigh:

I think it would be better if warnings were not generated for multiple instances of the same key where one is missing a default value.  Furthermore, there is an asymmetry in the current behaviour where default values only appear in the json if they are discovered last. Consider input A and input B.

#### input A
```ts
const myKeyNoDefault = t('my.key');
const myKey = t('my.key', 'foobar');
```

#### output A (current behaviour)

```
  [read]    /path/to/file.ts
  [warning] Found same keys with different values: my.key
```

```json
{
    "my": {
        "key": "foobar"
    }
}
```

#### input B
```ts
const myKey = t('my.key', 'foobar');
const myKeyNoDefault = t('my.key');
```

#### output B (current behaviour)

```
  [read]    /path/to/file.ts
  [warning] Found same keys with different values: my.key
```

```json
{
    "my": {
        "key": ""
    }
}
```

I think it would be better if it didn't generate a warning, and that you always got the default value regardless of the order.

#### output B (changed in this PR)

```
  [read]    /path/to/file.ts
```

```json
{
    "my": {
        "key": "foobar"
    }
}
```

#### input C

Of course you should still get a warning for doing something like this.

```ts
const myKey = t('my.key', 'foobar');
const myKeyNoDefault = t('my.key');
const myKeyDifferentDefault = t('my.key', 'bazquux');
```

#### output C (behaviour unchanged)
```
  [read]    /path/to/file.ts
  [warning] Found same keys with different values: my.key
```

```json
{
    "my": {
        "key": "bazquux"
    }
}
```

No

### Checklist

### DIFF
--- a/dist/helpers.js
+++ b/dist/helpers.js
@@ -83,7 +83,15 @@ function dotPathToHash(entry) {var target = arguments.length > 1 && arguments[1]
   var lastSegment = segments[segments.length - 1];
   var oldValue = inner[lastSegment];
   if (oldValue !== undefined && oldValue !== newValue) {
-    conflict = (0, _typeof2["default"])(oldValue) !== (0, _typeof2["default"])(newValue) ? 'key' : 'value';
+    if ((0, _typeof2["default"])(oldValue) !== (0, _typeof2["default"])(newValue)) {
+      conflict = 'key';
+    } else if (oldValue !== '') {
+      if (newValue === '') {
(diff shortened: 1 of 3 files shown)


karellm:
Thanks, I published `4.2.0` with this.

Issue #361, opened 2021-06-22
Parser convert unicode escape sequences in JSON


Elindorath:
## 🐛 Bug Report

This is very similar to #227, but using json output:

<img width="1233" alt="Capture d’écran 2021-06-22 à 12 16 28" src="https://user-images.githubusercontent.com/6784793/122907604-9d50da00-d353-11eb-92a8-afb452ec9828.png">

The `\u00a0` sequences are converted in utf8 (I guess?) when writing back the files.

## To Reproduce

In order to reproduce, just add `\u00a0` to any of your strings in a translation file and run the parser configured to output json files. This is a unicode escape sequence for a non-breaking space character.

## Expected behavior

The unicode escape sequences must stay intact after parser did its job.

## Your Environment

- *runtime version*: node v12
- *i18next version*: 20.3.2
- *os*: Mac

karellm:
Fixed in `4.5.0`

PR #362, opened 2021-06-22
Support unicode escape sequence in json

Elindorath:

When using unicode escape sequences, the parser converts them to utf8. This PR should allow to preserve them.

Be aware that this PR will also convert all non-ASCII characters to its unicode escape sequences as well. Personally, I find it best to make these characters explicit, but let me know if you'd like to see this behavior as an option.
Moreover, note that the case is not preserved inside the unicode escape sequences, it will always be lowercased (`\u00A0` will become `\u00a0`).

Yes #361 

### Checklist

### DIFF
--- a/dist/transform.js
+++ b/dist/transform.js
@@ -294,6 +294,11 @@ i18nTransform = /*#__PURE__*/function (_Transform) {(0, _inherits2["default"])(i
 
       } else {
         text = JSON.stringify(contents, null, this.options.indentation) + '\n';
+        // Convert non-ASCII characters to unicode escape sequence
+        text = text.replace(
+        /[\u007F-\uFFFF]/g,
+        function (chr) {return "\\u".concat("0000".concat(chr.charCodeAt(0).toString(16)).substr(-4));});
+
       }
 
       if (this.options.lineEnding === 'auto') {
--- a/src/transform.js
+++ b/src/transform.js
@@ -294,6 +294,8 @@ export default class i18nTransform extends Transform {
       })
     } else {
       text = JSON.stringify(contents, null, this.options.indentation) + '\n'
+      // Convert non-ASCII characters to unicode escape sequence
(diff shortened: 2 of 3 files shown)


karellm:
Sorry for the late answer, I had missed this one. Thanks for the PR!

kobrynsky:
Hi there!
shouldn't it be optional parameter? In my case I would like to have no escaped characters.

It currently converts it from this:
image
to this:
image

To work as I want it to work i commented code from this pull request:
image

But there is a possibility that I don't fully understand how I should handle it xD

I can make pull request in my free time to make this optional if you would accept it 😄 

Either way, this package is awesome. Thank you for your work.

damien-monni:
+1 on making this optional. My JSON file for french translation is unreadable now, so I have to stick with v4.2.0. 😅

image

karellm:
That's indeed a negligence on my part, I should have better tested this before merging the PR. I'm not sure if it should be optional or handled differently altogether. I will look into it later today and keep you posted.

Vovan-VE:
Forced escapes made cyrillic translation files to consist of escapes for about 99%, and so made files completely not user-friendly. This feature either should be optional or it should be in next major version as BC break feature.

Vovan-VE:
Probably, unicode properties `\p{...}` should be used at least to achieve original #361 goal.

bodryi:
Good day!
Tried to use this library today. All Latin characters are extracted correctly, but all Cyrillic ones are converted to Unicode escaped values. Had no idea why this happens, but then I looked source code and found, that changes were made in this PR.

Looking forward for this issue to be solved!

image_2021-08-03_13-44-55

Vovan-VE:
@bodryi , this issue was introduced in `4.3.0`, so you can use version `~4.2`.

bodryi:
Already downgraded version to 4.2.0 and it worked like a charm.

@Vovan-VE, thanks for advice!

karellm:
@Elindorath After looking a little more into this issue, I don't think this PR should make it to this project. I understand you case for preserving unicode escape sequence and if you want to submit a PR that does just that, I will look at it!

Everyone else, `4.4.0` is back on track!

PR #381, opened 2021-07-27
Extract tagged templates

### LINKED ISSUE #376: React - handle `t` when used as a tagged template
## 🚀 Feature Proposal

For React, support extraction when `t` is used as a tagged template (outside of being used as a normal function).

## Motivation

`t`, when used as a tagged template, offers a more concise syntax.

## Example

```tsx
const { t } = useTranslation();
<div>{t`Language`}<div>
```

kosciolek:

Fixes #376, more concise syntax.

Yes #376

### Checklist

- [Y] only relevant code is changed (make a diff before you submit the PR)
- [Y] tests are included and pass (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [N] documentation is changed or added
- [Y] I ran `yarn build` to compile and prettify the code

### DIFF
--- a/dist/lexers/javascript-lexer.js
+++ b/dist/lexers/javascript-lexer.js
@@ -57,6 +57,10 @@ JavascriptLexer = /*#__PURE__*/function (_BaseLexer) {(0, _inherits2["default"])
 
         parseCommentNode(keys, node, content);
 
+        if (node.kind === ts.SyntaxKind.TaggedTemplateExpression) {
+          entry = _this4.taggedTemplateExpressionExtractor.call(_this4, node);
+        }
+
         if (node.kind === ts.SyntaxKind.CallExpression) {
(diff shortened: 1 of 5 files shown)


kosciolek:
**Should any docs be added here?** I don't see any section should be notified of this change.

Tests will be added in a moment.

kosciolek:
Tests are added.

Currently it:
* **Extracts only non-interpolated templates.** `t\`some-key\``
* **Throws a warning on interpolated templates.** `t\`some-k${someVar}ey\`` -> `'A key that is a template string must not have any interpolations.'`

It sounds fine to me since this is just a more concise syntax for basic cases, but I've seen you're using comments to suggest values of dynamic keys. Should I try to implement it as well?

karellm:
Thanks for the PR, everything looks good to me.

Issue #305, opened 2021-04-30
Doesn't support array of strings passed to t function


tschanz-pcv:
## 🐛 Bug Report

In a `.tsx` file I have the following: `t(['flavor:welcome', 'welcome'])`
The array syntax can be found in the official docs, even with an example: https://www.i18next.com/overview/api#t

This syntax produces a warning `Key is not a string literal` and is ignored. The workaround with the comments is quite tedious and prone to be forgotten.

## To Reproduce

```tsx
import { FunctionComponent } from 'react';
import { useTranslation } from 'react-i18next';

const Demo: FunctionComponent = () => {
  const { t } = useTranslation();
  
  return (
    <span>{t(['flavor:welcome', 'welcome'])}</span>
  );
};
```

## Expected behavior

I would expect the array syntax to be supported.

## Your Environment

- *runtime version*: Node v14.16.1
- *i18next version*: 20.2.1 (i18next-parser: 3.11.0, react-i18next: 11.8.13)
- *os*: Mac

karellm:
Indeed, feel free to submit a PR!

karellm:
Fixed from `4.6.0`

PR #389, opened 2021-08-04
fix: added support for string arrays as an argument to the useTranslation

sarakusha:

added support for string arrays as an argument to the useTranslation

Yes/No #305

### Checklist

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -4,7 +4,7 @@ import * as ts from 'typescript'
 export default class JavascriptLexer extends BaseLexer {
   constructor(options = {}) {
     super(options)
-    this.namespaceSeparator = options.namespaceSeparator;
+
     this.callPattern = '(?<=^|\\s|\\.)' + this.functionPattern() + '\\(.*\\)'
     this.functions = options.functions || ['t']
     this.attr = options.attr || 'i18nKey'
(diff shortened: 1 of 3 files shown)


CarrotIronfoundersson:
This fix indeed solves the issue with array of namespaces as argument for useTranslation hook. I applied the change manually and tested it my project environment!

AlphaByte02:
This resolve also the #220 that was lost.

You probably need to add the test case for this PR to be valid and also is a good thing for the future.

karellm:
@sarakusha Thanks for the PR, can you please add a test to document this change? It is a requirement for me to merge a PR.

simonasdev:
would also love this feature

sarakusha:
> @sarakusha Thanks for the PR, can you please add a test to document this change? It is a requirement for me to merge a PR.

Actually, I practically didn’t do anything new. When I wanted to add functionality, it turned out that everything was implemented for `withTranslation`. I just changed the condition so that it all worked for `useTranslation`

karellm:
@sarakusha I understand but maybe the feature wasn't properly tested before. It doesn't hurt to add a test, see it as a way to document the change.

sarakusha:
I looked at how the parser works and found that it does not know how to handle constructs like `t("namespace:key")`, but it recognizes `t("key", { ns: namespace })`
I will create a new PR

sarakusha:
I had to add the passing of `options` from the `Parser` to `Lexer`, since the namespace separator is set in the `Transform` and passed to the `Parser`, but there was no way to get it in the `Lexer`

karellm:
@sarakusha I don't think this logic belongs in the Lexer. The Lexer is only extracting keys (whether it contains a namespace or not). Then the transform is in charge of making sense what is in the key. You can see the code here: https://github.com/i18next/i18next-parser/blob/master/src/transform.js#L103

That's the reason why Lexers don't need to know about the `namespaceSeparator`, it's not their concern. If you want to test for namespace handling, it should happen at the parser level, not the lexer level.

karellm:
Thanks. Deployed as `4.6.0`

PR #402, opened 2021-08-27
Remove unused 'colors' library

fredj:

The `colors` library is imported but not used. This pull request removes it.

No

### Checklist


karellm:
Yes it is: https://github.com/i18next/i18next-parser/blob/master/bin/cli.js#L35

PR #409, opened 2021-09-04
chore(ci): use setup-node native cache and major version of actions

mmorel-35:

HI ! This PR is to use the native cache of setup-node@v2 and use major versions of setup-node and codecov-action
It will avoid to have a PR for every new minor or patch version of those.

No

### Checklist


karellm:
Thanks for the PR!

PR #413, opened 2021-09-10
Escape non-printable Unicode characters

Vovan-VE:
Revert d3858b6dd1399cf9bf2fa6c36de93f7bab693380

Changed #362 to escape only non-printable Unicode character.

Enhanced version of #362. Escaping only Unicode Separators `\p{Z}`, Control `\p{Cc}` and Format `\p{Cf}` characters ([doc](https://unicode.org/reports/tr18/#General_Category_Property)).

See also: [node.green](https://node.green/#ES2018-features--RegExp-Unicode-Property-Escapes), [caniuse](https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes).

Tested also in proprietary project with Cyrillic translation files.
@kobrynsky, @damien-monni, @bodryi (menthioned from #362), feel free to test it in your projects too.

Yes #361

### Checklist

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -294,6 +294,12 @@ export default class i18nTransform extends Transform {
       })
     } else {
       text = JSON.stringify(contents, null, this.options.indentation) + '\n'
+      // Convert non-printable Unicode characters to unicode escape sequence
+      // https://unicode.org/reports/tr18/#General_Category_Property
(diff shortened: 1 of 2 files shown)


karellm:
Thanks for the PR. I have tested it quickly and it seems to work. I will deploy this as `4.5.0`

Issue #400, opened 2021-08-26
Unexpected sort behavior between JSON keys at the same level


Elindorath:
## 🐛 Bug Report

Sorting is done on the full JSON path rather than directly on the keys of JSON objects. For example, the sort will compare "header.title" against "header_confirmation.title" (and other combinations) resulting on putting the `header` object after the `header_confirmation` one.

## To Reproduce

Given a code using 4 keys as:
```js
t('header.title');
t('header.label');
t('header_confirmation.title');
t('header_confirmation.label');
```

The produced catalog is:
```json
{
  "header_confirmation": {
    "title": "Another title",
    "label": "Another label"
  },
  "header": {
    "title": "A title",
    "label": "A label"
  }
}
```

## Expected behavior

The produced catalog should be:
```json
{
  "header": {
    "title": "A title",
    "label": "A label"
  },
  "header_confirmation": {
    "title": "Another title",
    "label": "Another label"
  }
}
```

## Your Environment

- *runtime version*: node v14
- *i18next version*: 17.0.16
- *os*: Mac
- *i18next-parser version*: 4.0.1

karellm:
This is now fixed in `5.0.0` thanks to @Elindorath 's great work. Please read the [Changelog](https://github.com/i18next/i18next-parser/blob/master/docs/migration.md) and [migration guide](https://github.com/i18next/i18next-parser/blob/master/docs/migration.md) before updating.

Issue #169, opened 2019-09-07
Plural keys are not sorted properly (they appear at bottom)


hgezim:
- Is this a bug or a feature request?
Bug

- What version are you running (`0.x` or `1.x`)? 
1.0.0-beta34

- Which runner (cli, gulp...)?
cli

## What I'm trying to do?
Sort my keys.

## How am I doing it?
By setting `sort` to true in config.

Here is my config file:

```
// i18next-parser.config.js

module.exports = {  
    defaultNamespace: 'common',
    // Default namespace used in your i18next config
  
    defaultValue: '',
    // Default value to give to empty keys
  
    indentation: 4,
    // Indentation of the catalog files
    
    // see below for more details
    lexers: {
      js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
      jsx: ['JsxLexer'],
  
      ts: ['TypescriptLexer'],
      tsx: ['TypescriptLexer'],
  
      default: ['JavascriptLexer']
    },
  
    lineEnding: 'auto',
    // Control the line ending. See options at https://github.com/ryanve/eol
  
    locales: ['en', 'de', 'cs'],
    // An array of the locales in your applications
  
    namespaceSeparator: ':',
    // Namespace separator used in your translation keys
    // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.
  
    output: 'static/locales/$LOCALE/$NAMESPACE.json',
    // Supports $LOCALE and $NAMESPACE injection
    // Supports JSON (.json) and YAML (.yml) file formats
    // Where to write the locale files relative to process.cwd()
  
    input: undefined,
    // An array of globs that describe where to look for source files
    // relative to the location of the configuration file
  
    reactNamespace: false,
    // For react file, extract the defaultNamespace - https://react.i18next.com/components/translate-hoc.html
    // Ignored when parsing a `.jsx` file and namespace is extracted from that file.
  
    sort: true,
    // Whether or not to sort the catalog
  
    useKeysAsDefaultValue: false,
    // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
    // The option `defaultValue` will not work if this is set to true
  
    verbose: false
    // Display info about the parsing including some stats
  }
```

The command I'm running:

```
[...]

PR #418, opened 2021-09-15
Sort option as a function

simonasdev:

Needed a custom sort function to synchronise with a translation service.

No

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,16 +64,19 @@ npm install --save-dev i18next-parser
 [Gulp](http://gulpjs.com/) defines itself as the streaming build system. Put simply, it is like Grunt, but performant and elegant.
 
 ```javascript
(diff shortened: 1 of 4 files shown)


karellm:
@simonasdev Thanks for the PR. Please note that the sort logic is broken as you can see in #400 and #169 . Would you be interested in tackling this too?

On a side note, can you please remove the style changes to the README?

simonasdev:
Removed prettier changes to README. But I wouldn't call the current sort broken, it's just how `localeCompare` works - albeit unexpectedly, but consistently. In my case I just needed a simple ASCII sort, as in the test example I've made.

Unless you would consider making ASCII sort the default?

karellm:
Thanks for the PR, it is now part of `4.7.0`

Issue #419, opened 2021-09-16
[TypeScript lexer] Template literals cannot be used as `defaultValue` as a second argument


mariyan-borisov:
## 🐛 Bug Report

Since long strings have a high chance of containing both single and double quotes, I prefer to use template literals to avoid escaping the quotes as follows:
```js
i18n.t('myKey', {defaultValue: `My value`});
```
This seems to work perfectly fine.
Today I found out that there's a shorter way to provide a `defaultValue`:
```js
i18n.t('myKey', `My value`);
```
This one doesn't work and results in the following:
```json
"myKey": ""
```
Using a string literal instead of a template literal seems to work fine.

## To Reproduce

1. Translate a key and provide `defaultValue` as a second argument via an untagged template literal:
```js
i18n.t('myKey', `My value`);
```
2. Run `i18next` (`i18next-parser/bin/cli.js`)

## Expected behavior

The generated JSON file(s) should contain the following:
```json
"myKey": "My value"
```

## Your Environment

- *runtime version*: node v14.17.6
- *i18next version*: 4.6.0
- *os*: Linux

karellm:
@mariyan-borisov You are making a valid point but this is not a bug but a feature. Template literal could contain variables that cannot be compiled. That said, I understand your use case and if you want to open a PR that exclude only template literal that actually contain a variable, you are welcome to do so!

karellm:
Thanks for the PR. This is deployed as `4.8.0`

PR #420, opened 2021-09-16
Add template literal support for `defaultValue` as a second argument (#419)

mariyan-borisov:

It adds template literal support for `defaultValue` as a second argument.

Yes #419 

### Checklist


karellm:
Thanks, this is published as `4.8.0`

PR #423, opened 2021-09-20
Add support for JSON v4

Elindorath:

Adding support for [JSON v4](https://www.i18next.com/misc/json-format#i-18-next-json-v4)

No

### Checklist


Elindorath:
It still lacks of ordinal support yet

Elindorath:
I personally use only this extractor but I saw there is also [i18next-scanner](https://github.com/i18next/i18next-scanner).
So I ask myself: what's the reason to maintain two packages with identical purpose? Wouldn't it be more efficient to deprecate one and redirect users to the other?

karellm:
@Elindorath Thanks for the PR, I will review in the next couple days. 

As to your question, I couldn't agree more. I reached out to the dev and he didn't seem to share our vision or at least he never reached back : https://github.com/i18next/i18next-scanner/issues/19. What I can say is that this project still is used extensively so I keep maintaining it. Also the last time I checked, both projects still had different surface area.

Elindorath:
Closes #432

test/locales/en/test_context_plural.json
```diff
@@ -1,5 +1,5 @@
 {
   "first": "first",
-  "first_context1_plural": "first context1 plural",
+  "first_context1_other": "first context1 plural",
   "first_context2_plural_2": "first context2 plural 2"
```
ahoisl, line 4: Not sure what the context2 is, but `plural_2` looks fishy with the new plural suffixes

src/helpers.js
```diff
@@ -117,20 +117,24 @@ function dotPathToHash(entry, target = {}, options = {}) {
...
  * key).
  */
-function mergeHashes(source, target, keepRemoved = false) {
+function mergeHashes(source, target, options = {}) {
   let old = {}
   let mergeCount = 0
   let pullCount = 0
   let oldCount = 0
+
+  const keepRemoved = options.keepRemoved || false
```
ahoisl, line 135: During testing I noticed the `keepRemoved` option is no longer respected. We have `keepRemoved: true` in a config file and these entries are removed, anyway.

Caused by following lines not correctly calling mergeHashes with options object:
* helpers.js:143
* transform.js:197

ahoisl (changes requested):
There's a regression with `keepRemoved`. With the `keepRemoved` handling fixed, the new version works great with the new v4 format 👍 

Thanks a lot for your PR!

src/helpers.js
karellm: Good catch thanks!

test/locales/en/test_context_plural.json
karellm: `context2` is just from the test. You are right that `plural_2` is weird, that said it doesn't even show in the test. @Elindorath  Could you change it to `first_context2_two` and update the test in parser.js:642 to see if the plural is properly picked up?

karellm (commented):
Thanks a lot for the PR, it looks really good. I've added couple small comments in addition to the ones from @ahoisl. Could you have a look?

docs/migration.md
```diff
@@ -1,3 +1,11 @@
+# Migrating from `3.x` to `4.x`
```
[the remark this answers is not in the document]
Elindorath: Done

src/helpers.js
Elindorath: Done and added two tests for it

src/helpers.js
```diff
@@ -203,4 +211,14 @@ function transferValues(source, target) {
   }
 }
 
+function findRelatedPluralKey(rawKey, source) {
```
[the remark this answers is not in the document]
Elindorath: All done. I changed `!!source[`${rawKey}${suffix}`]` by `true` though: it didn't make sense to consider the value of the key as part of the result, only presence is relevant here

Elindorath:
I'd like to point out that in the current state, tests are dependent of the `Intl` API implementation in Node (Even if it not officially supported, try to run the tests in Node v10 to convince yourself). Ideally, the `Intl` API should be mocked to remove that dependency but I don't know if its required at this point and if so, the right way to do it

Elindorath:
Another thought, why keep track of the build artifacts? It would make more sense to me to build only when releasing on Npm. It could be made in another PR though

ahoisl:
> 
> 
> During developing the test `` copies `source` nested keys to `target` regardless of presence when `keepRemoved` is enabled ``, I noticed the `oldCount` property returned by the `mergeHashes` function was off when working with nested keys. Not sure if I did the fix well though ([src/helpers.js:146](https://github.com/i18next/i18next-parser/pull/423/files#diff-49ba0c85360693b8e89ae0e0428284a55db749f08f2b9993d26910def646e02fR146))

Looks good to me. Seems the `oldCount` should be populated regardless of the `keepRemoved` option 👍

ahoisl (approved):
Tested changes again, looks good to me 👏 👍 Thanks again!

ahoisl:
> As pointed out in #435, key sorting will be rather strange after merging this PR.
> ...
> Let me know if you want this to be included in the scope of this PR or not

Frankly, I do not experience anything strange about the sorting. It's actually better than before with the "_plural", because now all keys have a suffix. We only use "zero" and "other" and it works fine for us.

So for what it's worth, I would love to see this PR integrated so we can finally switch to v4 😄

test/locales/en/test_context_plural.json
karellm: It starts to smell like a bug. That's also why I was curious if changing it to the new plural form changed the outcome of the test. @Elindorath would you mind testing?

karellm:
Thanks a lot for the great work! I've merged and deployed the code as `5.0`.

@Elindorath I would greatly appreciate a PR that solves the sorting if you have the bandwidth for it. I also like your idea of removing the build artifacts.

PR #442, opened 2021-10-04
Enhanced publication

Elindorath:

Following [this question](https://github.com/i18next/i18next-parser/pull/423#issuecomment-931301291) I asked, here is a PR that eliminates build artifacts as well as many other improvement attempts. I tried to keep a clear commit history but here are some key changes:
- Build artifacts (`dist/` directory) are now ignored by git
- Published files are limited to build artifacts only (`npm publish --dry-run` to see the list)
- Prettier now run as a pre-commit hook and a pre-publish hook (not required anymore in PR for contributors)
- Manual tests previously required (for CLI, Gulp & Broccoli) are now running on CI (the way it was done is questionable though)

I didn't edit the version or edit the changelog, but this could be considered breaking changes for people directly requiring from `src/`:
```js
import { getPluralSuffixPosition } from 'i18next-parser/src/helpers.js'
```

No

### Checklist


karellm:
@Elindorath Thanks for the PR. I really like that you automated the manual test! The code looks good too. My only concern is that CI doesn't seem to pass at the moment. Do you know what is going on?

Elindorath:
> @Elindorath Thanks for the PR. I really like that you automated the manual test! The code looks good too. My only concern is that CI doesn't seem to pass at the moment. Do you know what is going on?

Yes! As I mentioned [here](https://github.com/i18next/i18next-parser/pull/423#issuecomment-931289762), tests that rely on the `Intl` dependency are subject to potential breaks across node versions. For this particular case, Node v14.17.0 introduced [an update of its `ICU` deps](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#other-notable-changes) to v68.2, which contains [an update of the CLDR v38](https://icu.unicode.org/download/68). [This version](https://cldr.unicode.org/index/downloads/cldr-38#h.7ahoxbsngtxp) added a plural form to french.
If I have the time, I will look for a proper way of mocking the `Intl` API, but for now, I fixed it by just adding the new plural form to the expected result

Elindorath:
> @Elindorath Thanks for the PR. I really like that you automated the manual test! The code looks good too. My only concern is that CI doesn't seem to pass at the moment. Do you know what is going on?

Hmmm yes, but it might be tricky to solve! As I mentioned [here](https://github.com/i18next/i18next-parser/pull/423#issuecomment-931289762), tests that rely on the `Intl` dependency are subject to potential breaks across node versions. For this particular case, Node v14.17.0 introduced [an update of its `ICU` deps](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V14.md#other-notable-changes) to v68.2, which contains [an update of the CLDR v38](https://icu.unicode.org/download/68). [This version](https://cldr.unicode.org/index/downloads/cldr-38#h.7ahoxbsngtxp) added a plural form to french.
But this change was not backported on Node v12...
We should definitively mock the `Intl` API to have consistent results, but for the previous manual tests, I don't know how to do that as they spawn another process.
I think I'll look for a different way to test those

Elindorath:
Despite the green tests, please wait a bit before merging. I try to change the way broccoli plugin and CLI are tested. I'll write back when its ready

Elindorath:
I think it is ready to merged in the current shape.
I only changed the broccoli test.
The CLI will have to wait a following PR to be tested programatically as it requires some more work.

.github/workflows/node.js.yml
```diff
@@ -17,10 +17,10 @@ jobs:
       with:
         cache: npm
```
mmorel-35, line 18: Hi! If you change from npm to yarn use it also in the cache system 
```suggestion
        cache: yarn
```

PR #443, opened 2021-10-05
Add @babel/runtime as deps

raon0211:

`i18next-parser` is requiring `@babel/runtime` in its output file, but `@babel/runtime` is not specified in its dependencies. When used with strict package managers such as pnpm or Yarn 2+, it causes the program to break. ([relevant article](https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies))

https://github.com/i18next/i18next-parser/blob/9d7f24d05fc28f5e55327d14c9ed1e7d1beb1950/dist/transform.js#L1

I added `@babel/runtime` as a regular dependency to the package.

No

### Checklist


karellm:
@raon0211 Thanks for the PR. Would you mind rebasing from master?

Issue #267, opened 2021-03-26
Default value as the single source of truth (overwrite existing values in the json file)


jgoux:
Hello,

## 🚀 Feature Proposal

I'd like to have an option to make the default values defined in the `t` function to overwrite existing value in the produced json file.

## Motivation

Right now the parser is detecting new keys and their default values used across the codebase and add them to the json file. But if we change the default value of a key in the codebase, the change isn't reflected in the json file. I think it would be better to reflect the change, or at least give an option to allow it.

## Example

```
// a new config option
{
  overwriteDefaultValues: true
}
```

karellm:
The parser doesn't currently keep a record of the defaultValues so it has no way to know if it has changed. 

Such an option would only allow all translations to be replaced by the default values. And I'm not really see value in that, even though I'm not opposed to it. 

If you are willing to implement a proper solution that tracks defaultValues and a `resetDefaultValueOnChange` option, I will happily review and merge the PR!

jgoux:
I didn't have tracking default values in mind, but something like :
- Read the existing json
- For each "t" function detected, if the key already exists in the json and there is a default value defined in "t", replace the key's value in the json.

I guess most of this workflow is already implemented so you can add new keys to the existing json?

karellm:
I understood your proposal but I'm not keen to add it. It would overwrite all translations with the default value, even the ones that are still good. How is that a good thing?

jgoux:
I'm not sure I understand this :

> It would overwrite all translations with the default value, even the ones that are still good

What do you mean by still good? Maybe our usage of this library is not regular but as we use our default values as the single source of truth to generate a default en.json file, with the change I'm suggesting nothing would be lost.

The keys that already exists but which values weren't modified won't be modified either in our en.json file. And the modified ones will be replaced but only in our default english file. Once it's done we use a translation service like lokalise that will detect that values have been modified for these keys and notify the translators so they can update the other languages.

What am I missing? It seems a reasonable workflow to me, our code is our only source of truth for our default i18n file and I don't expect anyone in the team to deal with the en.json file directly. And all the other translations files are dealt with on the external service outside of our dev workflow.

karellm:
I'm not used to rely on the default value to populate my translation files but it now makes sense to me. If you want to make a PR, I will happily review it.

worldlee78:
I also use the defaultValue in this way. Perhaps I misunderstood what the default value was meant to do if this workflow isn't how you're using it?

[...]

PR #451, opened 2021-10-13
Add resetDefaultValueLocale option

lars-mortenson:

Add a feature related to resetting translations for a key if the default value changes.

Yes #267 

### Checklist

- [ X] only relevant code is changed (make a diff before you submit the PR)
- [ X] tests are included and pass: `yarn test` (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [ X] documentation is changed or added

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ module.exports = {
   failOnUpdate: false,
   // Exit with an exit code of 1 when translations are updated (for CI purpose)
 
-  customValueTemplate: null
+  customValueTemplate: null,
   // If you wish to customize the value output the value as an object, you can set your own format.
   // ${defaultValue} is the default value you set in your translation function.
   // Any other custom property will be automatically extracted.
@@ -210,6 +210,12 @@ module.exports = {
   //   message: "${defaultValue}",
   //   description: "${maxLength}", // t('my-key', {maxLength: 150})
   // }
+
+  resetDefaultValueLocale: null
+  // The locale to compare with default values to determine whether a default value has been changed.
+  // If this is set and a default value differs from a translation in the specified locale, all entries
+  // for that key across locales are reset to the default value, and existing translations are moved to
+  // the `_old` file.
 }
 ```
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -273,8 +282,8 @@ Typescript is supported via Javascript and Jsx lexers. If you are using Javascri
   handlebars: [
     {
       lexer: 'HandlebarsLexer',
-      functions: ['t'] // Array of functions to match
-    }
+      functions: ['t'], // Array of functions to match
+    },
```
karellm, line 286: Please make sure you only commit changes relevant to what you worked on. I don't want to merge style changes.

karellm (commented):
Thanks for the PR, it looks good. My only concern is that I would like more test at the parser level. I don't like the code coverage reduction. Could you add some while you address my other comment about the README?

README.md
lars-mortenson: > Thanks for the PR, it looks good. My only concern is that I would like more test at the parser level. I don't like the code coverage reduction. Could you add some while you address my other comment about the README?

No problem, I should have an update shortly.

lars-mortenson:
> Thanks for the PR, it looks good. My only concern is that I would like more test at the parser level. I don't like the code coverage reduction. Could you add some while you address my other comment about the README?

No problem, I should have an update shortly.

README.md
lars-mortenson: Just a heads up, I think all those additional changes were made automatically on commit because of this line: https://github.com/i18next/i18next-parser/blob/master/.lintstagedrc.js#L2

So, if other submitters change the README.md I suspect you'll see the same changes occur.

smolster:
@karellm any update on this PR? Looks like the changes you requested have been addressed.

lars-mortenson:
Hey @karellm, any remaining issues with these changes?

karellm:
Sorry for the delay. Thanks for the great work, I'm deploying it as `5.2.0`

PR #471, opened 2021-11-03
Add `fail-on-update` option

fabien0102:

To be able to check if somebody forgot to extract the translations on CI, we need a way to check if everything is up-to-date.

This `fail-on-update` option solve this issue, exiting with code 1 if something was generated.

No

### Checklist


karellm:
Thanks for the PR. This is deployed as `5.1.0`

PR #506, opened 2022-01-17
jsx-lexer: Don't extract a variable identifier in i18nKey as key

raine:

Hello,

Updated to latest i18next-parser in our largeish project. There are few instances where we have `<Trans>` component with a variable as `i18nKey` for whatever reason. The latest i18next-parser started extracting those variables like so:

```
 <Trans
  i18nKey={titleTextKey}
  components={{
    rating_type: productRating,
  }}
/>
```

→ `"titleTextKey": ""`

Does not seem like the correct behavior for me hence the fix but please comment. Thanks!

Not sure.

### Checklist


kirillku (approved):

karellm:
Thanks for the PR. I believe a more consistent fix would be to throw a warning. You can see [here](https://github.com/i18next/i18next-parser/blob/92e7b4164eb2ca3bbb6db2a1442296f9852f5e3c/test/parser.test.js#L2351) and [here](https://github.com/i18next/i18next-parser/blob/38c0d9beea956472bbeb519b5678ea27a2b452e6/src/lexers/javascript-lexer.js#L147).

Would you mind adding the warning and a test?

raine:
@karellm Thanks for the input. How does it look now? https://github.com/i18next/i18next-parser/pull/506/commits/94536c82ed80eaf05e1f4bd6a8bf54378d1be6cb

PR #512, opened 2022-02-23
Set custom namespace functions

sk-idwell:
When you need to use a custom hook for translations

I'm using a custom hook and unable to extract namespaces without writing a custom lexer

No

### Checklist


karellm:
@sk-idwell This PR needs proper testing. Would you mind adding them?

sk-idwell:
@karellm Added a test

karellm:
This is deployed as `6.1.0`

Issue #489, opened 2021-12-09
Fail on update does not work with multiple namespaces


AnotherHermit:
## 🐛 Bug Report

When using multiple namespaces along with the "--fail-on-update" flag the result is always flagged as updated.
The issue seems to be the `countWithPlurals` that is used to check how many keys were added to a specific namespace. Since the `countWithPlurals` is set for the whole locale, not the specific namespace, the `addCount` for the namespace is wrong.
This can also be seen when the `verbose` setting is enabled, then the added keys will list the total number of keys for the locale minus the keys added in the specific namespace.

## To Reproduce

1. Use a setup with multiple namespaces with different keys in the different namespaces
2. Extract the keys once
3. Enable --fail-on-update (and verbose to see that the added keys is not 0)
4. Output will be flagged as having updates, even though output is the same

## Expected behavior

Fail on update should not always fail when run in environments with multiple namespaces

## Your Environment

- *runtime version*: node 16.5.0
- *i18next version*: 21.5.4
- *i18next-parser version*: 5.3.0
- *os*: Linux (WSL)

riccardo-forina:
I'm hitting this too.

karellm:
I'll happily review a PR if you want to take a spin at fixing this.

karellm:
This is likely fixed in `6.2.0`. Please not that `unique count` now includes plurals since it makes more sense with the new plural form.

alk-mandrianarijaona:
@karellm unfortunately upgrading to 6.2.0 does not solve my issue yet but I was not able to reproduce it in a new project with the same config as my private project. I will reopen later if I can find a way to reproduce it. Thanks 🙏🏼

PR #546, opened 2022-04-02
fix: count keys for each namespace instead of global

Laityned:

When using the verbose option, keys are counted globally while presented as if it is per file. This merge request make sure that statistics are given for each namespace

Yes,
As a side effect of the incorrect count for the number of unique/plural keys the fail-on-update bug as described in #489 is not working correctly. This should now be fixed too.

### Checklist

- [x ] only relevant code is changed (make a diff before you submit the PR)
- [ x] tests are included and pass: `yarn test` (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [ x] documentation is changed or added

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -148,8 +148,8 @@ export default class i18nTransform extends Transform {
       const catalog = {}
       const resetAndFlag = this.options.resetDefaultValueLocale === locale
 
-      let countWithPlurals = 0
-      let uniqueCount = this.entries.length
+      let countWithPlurals = {}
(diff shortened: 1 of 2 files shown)


karellm:
Thanks for the PR, I've merged it but it introduces new bugs since now duplicates are not taken into account. I will fix that and deploy a new version soon.

karellm:
This is fixed in `6.2.0`. Please not that `unique count` now includes plurals since it makes more sense with the new plural form.

Issue #495, opened 2021-12-23
Detect Reused Keys


amandapouget:
## 🚀 Feature Proposal

Allow detect key reuse and option to warn or fail on reuse.
Example: dev writes `t('foobar')` in two different files ultimately writing translations to the same JSON.
With config `warnOnDuplicates: true`, parser would log something like, `Key 'foobar' is reused 2 times.`
With config `failOnWarnings`, this situation would cause an exit 1.

## Motivation

Accidental key reuse results in wrong translations. The parser should be able to detect and prevent this scenario.
It happens especially in cases where the key is used as default value.

For example, `t('Cancel')` in one place might mean "Decide not to submit a form", while in another place, it might mean "End your subscription". Forcing developers to use unique keys prevents accidental mistranslations.

## Example

// Foo.js
```js
t('Foobar')
```
// Bar.js
```js
t('Foobar')
```
=> 'Warning: key 'Foobar' is reused 2 times.'

karellm:
A warning is thrown if there is a key use twice with a conflict on the value. What you are proposition is not yet implemented though.

PR #549, opened 2022-04-06
Detect reused keys

andrewRichardson:

Allow detect key reuse and option to warn or fail on reuse.
Example: dev writes `t('foobar')` in two different files ultimately writing translations to the same JSON.
With config `warnOnDuplicates: true`, parser now warns appropriately, and counts the number of warnings like so: `. Warnings:   XXX warnings encountered` to give a general idea of the number of reused keys.
With config `failOnWarnings`, this situation would cause an exit 1.

Yes #495 

### Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -8,7 +8,7 @@ import path from 'path'
 import sort from 'gulp-sort'
 import vfs from 'vinyl-fs'
 
-import i18nTransform from '../dist/transform.js'
+import i18nTransform, { KEY_REUSE_WARNING } from '../dist/transform.js'
 ;(async () => {
   const pkg = JSON.parse(
     await fsp.readFile(new URL('../package.json', import.meta.url), 'utf-8')
@@ -129,7 +129,7 @@ import i18nTransform from '../dist/transform.js'
   }
 
   var count = 0
-  var warnings = 0
+  let warningMap = {}
 
   vfs
     .src(globs)
@@ -157,15 +157,28 @@ import i18nTransform from '../dist/transform.js'
           if (!program.opts().silent) {
             console.log('  [warning] '.yellow + message)
           }
-          warnings++
+          if (message.startsWith(i18nTransform.KEY_REUSE_WARNING)) {
+            const key = message.substr(KEY_REUSE_WARNING.length)
+            const previousValue = warningMap[key] ?? 1
+            warningMap[key] = previousValue + 1
+          }
         })
         .on('finish', function () {
           if (!program.opts().silent) {
             console.log()
             console.log('  Stats:  '.cyan + count + ' files were parsed')
-            if (config.warnOnDuplicates && warnings > 0) {
+            if (config.warnOnDuplicates && Object.keys(warningMap).length > 0) {
+              var keys = []
+              for (var key in warningMap) {
+                keys.push([key, warningMap[key]])
+              }
+
+              keys.sort(function (a, b) {
+                return b[1] - a[1]
+              })
+              console.log(keys)
               console.log(
-                '  Warnings:  '.yellow + warnings + ' warnings encountered'
+                '  Key Reuse Warning:  '.yellow + keys.length + ' keys reused'
               )
             }
           }
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -75,6 +75,8 @@ import i18nTransform, { KEY_REUSE_WARNING } from '../dist/transform.js'
     program.opts().failOnUpdate || config.failOnUpdate || false
   config.silent =
     program.opts().silent || program.opts().getKeyDuplicates || false
+  config.warnOnKeyDuplicates =
+    program.opts().getKeyDuplicates || config.warnOnKeyDuplicates || false
 
   let args = program.args || []
   let globs
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -165,7 +165,7 @@ import i18nTransform, { KEY_REUSE_WARNING } from '../dist/transform.js'
           }
           if (message.startsWith(KEY_REUSE_WARNING)) {
             const key = message.substr(KEY_REUSE_WARNING.length)
-            const previousValue = duplicateKeys[key] ?? 1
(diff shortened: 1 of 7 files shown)


andrewRichardson:
@karellm

bin/cli.js
```diff
@@ -156,11 +157,30 @@ import i18nTransform from '../dist/transform.js'
...
+            const previousValue = warningMap[key] ?? 1
+            warningMap[key] = previousValue + 1
+          }
         })
         .on('finish', function () {
           if (!program.opts().silent) {
             console.log()
             console.log('  Stats:  '.cyan + count + ' files were parsed')
+            if (config.warnOnDuplicates && Object.keys(warningMap).length > 0) {
+              var keys = []
```
karellm, line 171: I know there are inconsistencies in this file but I'd prefer we stick with `let/const`. Any reason you went for `var`?

bin/cli.js
```diff
@@ -129,6 +129,7 @@ import i18nTransform from '../dist/transform.js'
   }
 
   var count = 0
+  let warningMap = {}
```
karellm, line 132: Since this is only used for duplicated keys I'd probably name this `duplicatedKeysCounts` or something similar.

bin/cli.js
```diff
@@ -156,11 +157,30 @@ import i18nTransform from '../dist/transform.js'
...
+            if (config.warnOnDuplicates && Object.keys(warningMap).length > 0) {
+              var keys = []
+              for (var key in warningMap) {
+                keys.push([key, warningMap[key]])
+              }
+
+              keys.sort(function (a, b) {
+                return b[1] - a[1]
+              })
+              console.log(keys)
```
karellm, line 179: This would warrant a test imo

karellm (commented):
Thanks for the PR, it looks good! I've added couple comments. My main concern is that tests are a bit light.

bin/cli.js
andrewRichardson: I went with `var` to match the `count` variable above it, but I can just switch both to `let` to fall more inline with the latest JS practices 👍🏼

andrewRichardson:
> @karellm  Thanks for the PR, it looks good! I've added couple comments. My main concern is that tests are a bit light.

I addressed your concerns including adding stronger tests.  Let me know if you have any other concerns! Thanks.

test/cli/cli.test.js
```diff
@@ -82,6 +82,12 @@ describe('CLI', function () {
     assert.deepEqual(frEmpty, {})
   })
 
+  it('works with `--get-key-duplicates` option', async () => {
+    const subprocess = await execaCommand('yarn test:cli --get-key-duplicates')
+
+    assert.include(subprocess.stdout, "[ [ 'seventh', 3 ] ]")
```
karellm, line 88: I'm confused why this would be `3` and not `2`. The file has only two occurrences of the key.

andrewRichardson: I went back and looked at how I counted the keys in my own personal implementation compared to the test translation file, and they are slightly different.  That meant that I was adding an extra "count" to each duplicate.  I fixed that in the cli function and it should be sensical now as it correctly counts `2` here etc.

test/helpers/dotPathToHash.test.js
```diff
@@ -96,6 +96,18 @@ describe('dotPathToHash helper function', () => {
     done()
   })
 
+  it('detects duplicate keys with the same value when warnOnDuplicate is true', (done) => {
+    const { target, duplicate, conflict } = dotPathToHash(
+      { keyWithNamespace: 'one.two.three' },
+      { one: { two: { three: '' } } },
+      { warnOnDuplicate: true }
```
karellm, line 103: The option is `warnOnKeyDuplicate` and not `warnOnDuplicate`. There is something flawed with the logic in this PR. This option shouldn't be needed to be passed to `dotPathToHash`. It should already return duplicates and conflicts without any extra option. This test is basically a duplicate of the one right above it.

src/transform.js
```diff
@@ -182,7 +185,7 @@ export default class i18nTransform extends Transform {
                 `new key already mapped to a string: ${entry.key}`
             )
           } else if (conflict === 'value') {
-            this.warn(`Found same keys with different values: ${entry.key}`)
+            this.warn(KEY_REUSE_WARNING + entry.key)
```
karellm, line 188: I believe what we want is a new warning for cases where `duplicate === true` and `conflict === false` which currently is not covered by the code. You should rename `KEY_REUSE_WARNING` to `KEY_CONFLICT_WARNING` and create a new warning for `KEY_REUSE_WARNING`.

karellm (commented):
Doing a last review I noticed inconsistencies with the code. Please review my two comments. I think the approach to solve this issue is not ideal and I've suggested an alternative.

test/helpers/dotPathToHash.test.js
andrewRichardson: Addressed in your other comment below

src/transform.js
andrewRichardson: Understood!  It seems like I ended up targeting only my use case and missed the bigger picture.  I reworked the code to use the `conflict = false` condition and everything seems to be working well, great suggestion.

andrewRichardson:
@karellm bump

amandapouget:
This feature seems really cool! Can we get to merge on it?

test/helpers/dotPathToHash.test.js
karellm: @andrewRichardson Is there a reason you left the test? It doesn't seem to test anything since you've removed `warnOnDuplicate`.

andrewRichardson: Ah, you are correct.  This was left by mistake, I just pushed a commit to remove it 👍🏼

andrewRichardson:
@karellm bump x2

andrewRichardson:
@karellm bump x3

andrewRichardson:
@karellm bump x4?

PR #581, opened 2022-05-29
Fix: run all test instead of single one

Laityned:

Test suite is adapted to run only a single test. Removing .only with make sure all test runs on `yarn test`

No

### Checklist


karellm:
Thanks for the catch!

PR #582, opened 2022-05-29
Fix: reset nested keys if default value is changed

Laityned:

the option resetDefaultValues is not working for nested keys. This PR fixes this bug in by changign the helper mergeHashes. 
Futhermore it includes tests for resetting nested keys in the mergeHashes helper and fully integration test of the parser for nested keys. 

Yes #529  

### Checklist


Laityned:
Done

karellm:
This was added to `6.5.0`

Issue #597, opened 2022-06-08
failOnUpdate fails when no update happens, when using namespaces


JulioC:
## 🐛 Bug Report

When using failsOnUpdates flag, the transformers emits an error even when no update is necessary. This only happens when using multiple namespaces.

It appears that the cause is that the namespaces won't be considered for the original number of tokens when calculating the change count (see below).

https://github.com/i18next/i18next-parser/blob/42b66c1dca5986f462825057e10db58222f5ddc0/src/transform.js#L278-L284

## To Reproduce

Consider the JS below (or any code you have)

```js
t('ns1:key1', 'Key 1');
t('ns2:key2', 'Key 1');
```

Now run the CLI against it to generate the JSON files, without `failOnUpdate`. This should run fine.

Finally, run the CLI once again, but this time add `--fail-on-update`. This will fail even without changes in JS source.

You can check that no update is necessary by add running the CLI without that flag once again.

Also, run with `--fail-on-update --verbose` to see the incorrect totals.

## Expected behavior

It should only emit an error when update is necessary

## Your Environment

- *runtime version*: i.e. node v16
- *i18next version*: i.e. 6.4.0

karellm:
@JulioC Could you make a PR with proper testing to fix this? Thanks for reporting.

JulioC:
@karellm I've added a new test in https://github.com/i18next/i18next-parser/pull/600, they should be failing as described in this thread.

---

As an additional note, I've also had some trouble running the CLI test suite on Windows because it's trying to run `yarn test:cli`, which has two commands sequentially run using `&&`. This is not valid in powershell nor in cmd.exe.

I suggest moving the `yarn -s build` part of that command to a `beforeAll` block for the CLI tests. This should also improve test speeds as it would be run only once.

Laityned:
Using the latest state of this repo, this issue can not be replicated. I wrote test cases directly in the parser.test.js as well. What I did found out is that this behaviour is probably related #634 . The number of added keys is wrong in case of an empty namespace.

PR #600, opened 2022-06-13
Improve tests for --fail-on-update CLI option

JulioC:

Add tests to check if the CLI runs without errors when `--fail-on-update` is used, but no update is required

No, but should be a starting point for fixing https://github.com/i18next/i18next-parser/issues/597

### Checklist


Laityned:
This is failing because of #634, not the issue you mentioned.

PR #620, opened 2022-07-22
Updated exports to fix pod install in RN

horne3754sg:

To fix an issue with React Native pod install failing because `package.json` is not included as an export  in `package.json`  
([see node package exports reference](https://nodejs.org/api/packages.html#exports))

No

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,12 @@
   "type": "module",
   "license": "MIT",
   "author": "Karel Ledru",
-  "exports": "./dist/index.js",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,7 +5,12 @@
   "type": "module",
   "license": "MIT",
   "author": "Karel Ledru",
-  "exports": "./dist/index.js",
+  "exports": {
+    ".": {
```
karellm, line 9: Is there a reason you don't simply do `".": "./dist/index.js"` ?

karellm (commented):
Thanks for the PR, I added a question.

karellm:
Closing by lack of activity. If you can point me to a proper ressource explaining why the change is needed, I'll happily review it again.

PR #626, opened 2022-08-18
Support custom yaml output options

bcentdev:

It would be nice to allow users to have their own settings for yaml output. In fact, I need it for a project :)

No

### Checklist


karellm:
Deployed as part of 6.6.0. Thanks!

PR #637, opened 2022-10-03
Fix: do not fail on JSX spread attribute in Trans component

### LINKED ISSUE #367: Cannot read property 'text' of undefined in jsx-lexer
## 🐛 Bug Report

Same behavior as https://github.com/i18next/i18next-parser/issues/199 where the spread operator generates an error using i18next CLI on `v4.1.1`

## To Reproduce

```
import React from 'react';
import PropTypes from 'prop-types';
import { withTranslation, Trans } from 'react-i18next';

// Refer to https://react.i18next.com/latest/trans-component#alternative-usage-components-array
const Component = (props) => {
  return <Trans {...props}>{props.children}</Trans>;
};

Component.propTypes = {
  children: PropTypes.node.isRequired
};

export default withTranslation()(Component);
```

Once I remove the `{...props}`, then this error disappears.

## Expected behavior

There should be no error.

## Your Environment

- *runtime version*: i.e. node v12.21
- *i18next version*: i.e. 19.5.3
- *os*: Mac
- *any other relevant information*: the workaround is to remove the spread operator.

dimazollo:

#367

Yes, it fixes #367

### Checklist

### DIFF
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -64,9 +64,9 @@ export default class JsxLexer extends JavascriptLexer {
   jsxExtractor(node, sourceText) {
     const tagNode = node.openingElement || node
 
-    const getPropValue = (node, tagName) => {
+    const getPropValue = (node, attributeName) => {
       const attribute = node.attributes.properties.find(
-        (attr) => attr.name.text === tagName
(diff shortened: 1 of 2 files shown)


karellm:
Deployed as part of 6.6.0. Thanks

PR #659, opened 2022-10-20
update example with correct lexer config variable name

Cielquan:
https://github.com/i18next/i18next-parser/blob/master/src/lexers/javascript-lexer.js#L10

The name of the lexer config variable does not match with the case in the code base. I fixed that.
See [here](https://github.com/i18next/i18next-parser/blob/master/src/lexers/javascript-lexer.js#L10) for the usage in code.

No

### Checklist


karellm:
Thanks!

Issue #664, opened 2022-10-24
Escaped string in JSX is not unescaped in defaultValue


yoo2001818:
## 🐛 Bug Report
In JSX, you can specify HTML entities (`&lt;`, `&#34;`, `&quot;`, etc) in JsxText.

```jsx
<Trans>
  &quot;Hello &amp; world!&#34;
</Trans>
// Would be parsed as... "Hello & world!"
```

These HTML entities are unescaped while compile time - so it would end up like this in the compiled code:

```js
React.createElement(Trans, null, "\"Hello & world!\"");
```

However, i18next-parser's JsxLexer doesn't unescape JsxText properly, making `defaultValue` to be `&quot;Hello &amp; world!&#34;` in the example. Similiar to #663, if `i18nKey` is specified in other ways, it can be patched by specifying [`shouldUnescape` to true](https://react.i18next.com/latest/trans-component#trans-props).

But, if `i18nKey` is not specified (or `shouldUnescape` is not applied), react-i18next never sees the escaped string because it would only reside in the raw source code - the transpiler effectively removes the escape from the transpiled code, making the translation key to differ between the generated file and the actual code.

This can be resolved by changing JsxLexer's `JsxText` rendering routine - https://github.com/i18next/i18next-parser/blob/master/src/lexers/jsx-lexer.js#L186-L191

```js
        if (child.kind === ts.SyntaxKind.JsxText) {
          return {
            type: 'text',
            content: unescape(child.text)
              .replace(/(^(\n|\r)\s*)|((\n|\r)\s*$)/g, '')
              .replace(/(\n|\r)\s*/g, ' '),
          };
```
Simply unescaping the text would resolve the problem. However, I'm not sure if I should send PR with a new dependency (e.g. [unescape](https://www.npmjs.com/package/unescape)), or bundle simple `unescape` function inside (https://github.com/i18next/react-i18next/blob/master/src/unescape.js). I'll be looking forward for your suggestions!

## To Reproduce
```jsx
<Trans>
  &quot;Hello &amp; world!&#34;
</Trans>
```
is parsed/generated as-is - `&quot;Hello &amp; world!&#34;` is the defaultValue / i18nKey.

## Expected behavior
`"Hello & world!"` should be the generated defaultValue / i18nKey.

## Your Environment

- *runtime version*: Node 14
- *i18next version*: 11.18.6
- *os*: Linux
- *any other relevant information*

nicegamer7:
Fixed as of #892.

[...]

Issue #264, opened 2021-03-03
Trans should not generate <1></1> for <br /> and <p>


tkrotoff:
## To Reproduce

```JSX
function Test() {
  const { t } = useTranslation();

  return (
    <Trans t={t}>
      hello
      <br />
      world
    </Trans>
  );
}
```

```
i18next Test.ts
```

i18next-parser generates:

```JSON
{
  "hello<1></1>world": "salut<1></1>monde",
}
```

i18next in debug mode reports:
```
i18next::translator: missingKey fr translation hello<br/>world hello<br/>world
```

## Expected behavior

i18next-parser should generate instead:

```js
{
  "hello<br/>world": "salut<br/>monde",
}
```

## More

`<p>` does not work too: i18next-parser should generate `hello<p>world</p>` instead of `hello<1>world</1>`.

Funny: `<small>` generates `hello<1>world</1>` and it works with i18next!!

There is definitely something wrong with i18next and i18next-parser formats: they should be the same.

## Your Environment

- *runtime version*: node 12, Firefox 86 & Chrome 88
- *i18next version*: i.e. 19.9.1
- *i18next-parser version*: 3.7.0
- *react-i18next version*: 11.8.8

karellm:
It might be related to https://github.com/i18next/i18next-parser/pull/254

I'm not using react and I'm mostly maintaining this project because many rely on it. If you want to submit a PR, I will review and merge it!

igorsantos07:
Yep, that seems to still be a bug under i18next-parser 4.3.0. It seems even a tad worse, since it seems to lose track of the element indexes?

```
Setting => transKeepBasicHtmlNodesFor: ['p', 'br', 'strong', 'b', 'i', 'em', 'u', 'tt', 'label']
i18next reports =>      [...] <1>Use <tt>0</tt> to disable</1>. 
i18next-parser writes => [...] <1>Use <1>0</1> to disable</1>.
```

karellm:
@igorsantos07 Yes ideally it should follow the settings from `i18next`.

[...]

PR #665, opened 2022-10-24
Fix tag converting behavior to match with react-i18next

yoo2001818:
This PR is continuation of #663. (You can disregard #663 if this PR gets merged first)

I've checked the difference between react-i18next and i18next-parser, and mainly there were 3 problems.

- `<br />` is converted to `<br />` (should be `<br/>`)
  - Subsequently, `<strong></strong>` should be converted to `<strong/>` in this case. This directly contradicts #254, so I think I need your decision whether to strictly follow react-i18next's behavior, or to allow some differences.
- Escaped string literal is not unescaped (#664)
- Finally, react-i18next only allows using `elementName` when only 1 string element is inside the node.
  For example, `<p>a{'b'}</p>` is converted to `<1>ab</1>` because there are two string elements inside `<p>`.
  `<p>{{ count }}</p>` is also converted to `<1>{{ count }}</1>` because the children is an object.
  - https://github.com/i18next/react-i18next/blob/master/src/Trans.js#L69

This resolves 1st and 3rd point to match with react-i18next's behavior.

I think #264 can be resolved if this PR gets merged, and `transSupportBasicHtmlNodes` becomes `true` by default. However, since that would be a breaking change, a proper documentation about `transSupportBasicHtmlNodes` could be enough to resolve it. I'll try to update the documentation if you want to.

No

### Checklist


mskrzypek:
@yoo2001818 `transSupportBasicHtmlNodes` is `true` by default, isn't it? https://react.i18next.com/latest/i18next-instance

yoo2001818:
> > I think #264 can be resolved if this PR gets merged, and transSupportBasicHtmlNodes becomes true by default.
> 
> @yoo2001818 `transSupportBasicHtmlNodes` is `true` by default, isn't it? https://react.i18next.com/latest/i18next-instance

No, it is `true` by default in i18next itself, but i18next-parser defaults to `false` (https://github.com/i18next/i18next-parser/blob/master/src/lexers/jsx-lexer.js#L10).

You have to match them, either by setting i18next's value to `false` (which will work perfectly for now), or setting i18next-parser's value to true. (which this PR tries to resolve discrepancies)

mskrzypek:
> You have to match them, either by setting i18next's value to `false` (which will work perfectly for now), or setting i18next-parser's value to `true`. (which this PR tries to resolve discrepancies)

As far as i can see, i18next-parser does not have `transSupportBasicHtmlNodes` setting itself. Correct me if I'm wrong, but It is derived from i18next settings, with default (fallback) to `false` indeed.

yoo2001818:
> > You have to match them, either by setting i18next's value to `false` (which will work perfectly for now), or setting i18next-parser's value to `true`. (which this PR tries to resolve discrepancies)
> 
> As far as i can see, i18next-parser does not have `transSupportBasicHtmlNodes` setting itself. Correct me if I'm wrong, but It is derived from i18next settings, with default (fallback) to `false` indeed.

i18next-parser does have `transSupportBasicHtmlNodes` setting, however it is undocumented for now.

You may use it like this:
```js
// i18next-parser.config.js
export default {
  lexers: {
    jsx: [{
      lexer: 'JsxLexer',
      transSupportBasicHtmlNodes: true,
    }],
  },
};
```

Furthermore, i18next-parser cannot read i18next's config, as i18next-parser statically analyzes the code without executing it. It doesn't actually care about i18next's configuration or its presence - it works by just reading `t` function call (or `<Trans>`) in the code. So it's the developer's job to match these config.

However, i18next itself, does set `transSupportBasicHtmlNodes` to `true` by default. What I've linked was for i18next-parser, which sets the config to `false` by default.

jaycle:
Bumped into this today and was so grateful to find not only it is reported but there's a PR already. Could this be looked at by a maintainer?

Issue #671, opened 2022-11-02
`defaultValue` should take precedence over `useKeysAsDefaultValue` if present


JonathanSun:
## 🚀 Feature Proposal

When `useKeysAsDefaultValue=true`, only apply the replacement if no `defaultValue` is specified.

## Motivation

Right now, there is no good way for `i18next-parser` to support code where some translations specify a default value and some want the key to be the default.  This is because the [current logic](https://github.com/i18next/i18next-parser/blob/345025312c825bfffc2909d44fa70165d81700d1/src/helpers.js#L62) always overwrites any existing default value even if it's specified.

## Example

For a codebase that has both usages:

```
t("Hello World!")
t("logout-message", "Goodbye World!");
```

be able to generate both keys at the same time:

```
"Hello World!": "Hello World!"
"logout-message": "Goodbye World!"
```

JonathanSun:
Consider #677 as a better alternative.

PR #672, opened 2022-11-03
Make defaultValue take precedence over useKeysAsDefaultValue if present

JonathanSun:

When useKeysAsDefaultValue=true, only apply the replacement if no defaultValue is specified.

Yes #671 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ export default {
 
   useKeysAsDefaultValue: false,
   // Whether to use the keys as the default value; ex. "Hello": "Hello", "World": "World"
-  // This option takes precedence over the `defaultValue` and `skipDefaultValues` options
+  // Will only take effect if `defaultValue` is not present or `skipDefaultValues` is true
   // You may also specify a function accepting the locale and namespace as arguments
 
   verbose: false,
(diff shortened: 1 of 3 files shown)


karellm:
@JonathanSun This is a breaking change so I'm very reluctant to add this to the project as it will introduce an unexpected behaviour for current users.

That said I think what you want to do make total sense. I believe a much better solution is to rely on the `defaultValue` option and pass it a function. The only argument missing to do what you want is the extracted default value. 

I believe we could remove `skipDefaultValues` and `useKeysAsDefaultValue` altogether and simplify the API while at it!

What do you think?

JonathanSun:
> @JonathanSun This is a breaking change so I'm very reluctant to add this to the project as it will introduce an unexpected behaviour for current users.
> 
> That said I think what you want to do make total sense. I believe a much better solution is to rely on the `defaultValue` option and pass it a function. The only argument missing to do what you want is the extracted default value.
> 
> I believe we could remove `skipDefaultValues` and `useKeysAsDefaultValue` altogether and simplify the API while at it!
> 
> What do you think?

Sure.  See https://github.com/i18next/i18next-parser/pull/676 for this suggestion.  It's still a breaking change, but I agree that it is a cleaner, more powerful option.

karellm:
Closing in favour of #676

PR #673, opened 2022-11-06
support .ts, .json and .yaml config

guoyunhe:

support .ts, .json and .yaml config

No

### Checklist


guoyunhe:
Here is one unit test failed. But I believe it is not related to my changes.

src/helpers.js
```diff
@@ -281,6 +286,37 @@ function makeDefaultSort(pluralSeparator) {
   }
 }
 
+async function esConfigLoader(filepath) {
+  return (await import(filepath)).default
+}
+
+async function tsConfigLoader(filepath) {
+  console.log(filepath)
```
karellm, line 294: `Console.log` :)

karellm (commented):
Thanks for the PR. Can you remove the console.log ?

src/helpers.js
guoyunhe: done

karellm:
Published as `7.0.0`

PR #676, opened 2022-11-08
Add entry value as a parameter to defaultValue function

JonathanSun:

This replaces the `skipDefaultValues` and `useKeysAsDefaultValues` with a more powerful option to use `value` itself as a function param.

Example:
```
defaultValue: (locale, namespace, key, value) => value ? value : `${key}`
```

Yes #677 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ export default {
 
   defaultValue: '',
   // Default value to give to keys with no value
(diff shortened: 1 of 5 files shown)


karellm (commented):
Thanks a lot, the PR looks great! I've added two small comments.

karellm:
Published as `7.0.0`

PR #685, opened 2022-11-09
feat: add config option to disable the non-printable utf8 escaping

cslecours:

In my project, when I run i18next-parser, I get the converted entities, however, my automated translation tool converts them back to the non-escaped sequence. 

This has become a little dance where as soon as I merge the translations I get another PR from the service to change it back.

Yes/No #000

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,10 @@ export default {
   // {
   //   lineWidth: -1,
   // }
+
+
+  escapeNonPrintableUnicodeCharacters: true
+  // If you do not want the parser to convert non printable characters to UTF-8 Character Sequences (ex : \\u00a0), put this to true.
 }
 ```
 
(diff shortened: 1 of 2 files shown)


cslecours:
This is only a proof of concept :  if the idea is approved, I can get the PR in a state where it's ready to merge, but I would not waste too much time if it's not the case :)

karellm:
Thanks for the PR. If you add a test for the new behaviour, I'll be happy to merge it.

brianmotherway:
Is there any update on this being merged? We're having issues with this affecting every commit we try to do in our codebase

karellm:
@brianmotherway The change is not tested. If you want to take over the PR and write a test for it, I'll review and merge it

cslecours:
@brianmotherway, if you write the tests that would be super helpful! 🙏

PR #691, opened 2022-11-14
feat: add support for `Translation` render prop

ggalletty:

Currently the parser doesn't take into account `<Translation>` namespace for children expression within the render prop.

```
<Translation ns="foo">{(t) => t("some_key")}</Translation>

// js lexer result will miss namespace `foo`, assigning keys to the default one
{ key: "some_key" }
```

There's no documentation related to translation components so am not sure if there's any to add here as well. 

Added test watch script to ease contributions. Figured I'd add that as well since it can benefit others. How I used it:
```
yarn test:watch test/lexers/jsx-lexer.test.js
```

Yes #702

### Checklist


karellm:
This is published as `7.3.0`

PR #692, opened 2022-11-14
fix: search file path when no --config option

guoyunhe:

Fix my previous PR #673 which removed default value for `--config` option but didn't search for config files with lilconfig correctly.

No

### Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -53,7 +53,7 @@ import i18nTransform from '../dist/transform.js'
 
   let config = {}
   try {
-    const result = await lilconfig(pkg.name, {
+    const configReader = lilconfig(pkg.name, {
       searchPlaces: [
         `${pkg.name}.config.js`,
         `${pkg.name}.config.json`,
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -67,7 +67,13 @@ import i18nTransform from '../dist/transform.js'
...
       },
-    }).load(program.opts().config)
+    })
+    let result
+    if (program.opts().config) {
+      result = await configReader.load(program.opts().config)
+    } else {
+      result = await configReader.search()
+    }
     config = result.config
```
karellm, line 77: This is still possibly problematic as if there is no config file this will fail (`result.config` is then `undefined`). I've pushed a fix as `7.0.2`

bin/cli.js
```diff
@@ -117,16 +123,19 @@ import i18nTransform from '../dist/transform.js'
       }
     }
 
+    let basePath = process.cwd()
```
karellm, line 126: Could you make a PR with only that part of the code? I've missed that in my fix.

karellm (commented):
Thanks, we've been at it at the same time and I've partially fixed it. See my comments.

karellm:
It is published as `7.0.3` Thanks for the PR

PR #701, opened 2022-11-21
feat: parse namespace from `t` type arguments

ggalletty:

It allows us to use type argument of `t` function (see [type definition](https://github.com/i18next/i18next/blob/master/index.d.ts#L881)) to set default namespace when parsing content.

This is particularly useful when you are passing `t` from component, for example

```
// component.tsx
export const UserDetail = () => {
  const { t } = useTranslation("user");
  const templateData = useMemo(() => getData(t), [t]);
  return <UserProfile help={t("help.tooltip")} content={templateData} />
};

// util.ts
export const getData = (t: TFunction<"user">) => ({
  greeting: t("user.greeting", "Welcome!"),
  intro: t("user.body"),
})
```

Even though the example above might not be the best use case, it should be good enough to explain how we can benefit from this.

In the case above, you'd normally end up with one translation in `user` namespace and the rest in the default namespace.
With the proposed changes, we would parse the function declarations to extract namespace from type arguments of `t` so that all parsed keys end up in `user` namespace.

What do you guys think about this? Would it be somewhat helpful? Have you found yourselves dealing with this kind of cases in the past?

Yes #703

### Checklist

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -74,6 +74,13 @@ export default class JavascriptLexer extends BaseLexer {
 
       parseCommentNode(keys, node, content)
 
+      if (
+        node.kind === ts.SyntaxKind.ArrowFunction ||
+        node.kind === ts.SyntaxKind.FunctionDeclaration
+      ) {
+        this.functionParamExtractor.call(this, node)
+      }
+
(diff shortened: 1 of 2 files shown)


karellm:
Thanks ! This is published as `7.3.0`

Issue #708, opened 2022-11-30
.mjs extension is no longer supported on the config file 


steveartfx:
## 💥 Regression Report

After upgrading from version 6.4.0 to 7.0.3 I get an error:

/node_modules/lilconfig/dist/index.js:84
        throw new Error(`No loader specified for extension "${ext}"`);

Error: No loader specified for extension ".mjs"

## Last working version

Worked up to version: 7.0.3

Stopped working in version: 7.0.3

## To Reproduce

```js
i18next -c './i18next-parser.config.mjs' 
```

## Expected behavior

.mjs file should still be a supported extension for the parser config file

```js
// Paste the results here
```

## Your Environment

- *runtime version*: i.e. node v16
- *i18next version*: i.e. ^21.6.14
- *os*: Mac
- *any other relevant information*

I found the issue is that .mjs is not white listed here, so hoping this is an easy fix:
bin/cli.js
![Screen Shot 2022-11-30 at 2 30 04 PM](https://user-images.githubusercontent.com/60870788/204926125-d9b559a0-6a0d-4535-86b4-20e278ee5ce9.png)

karellm:
This is fixed as of `7.2.0`

PR #733, opened 2022-12-16
restore .mjs as a config file extension

steveartfx:

Our projects are using .mjs as the config file extension and this stopped working since version 7 of parser. 

Yes/No #708 

### Checklist

- [ X] only relevant code is changed (make a diff before you submit the PR)
- [ X] tests are included and pass: `yarn test` (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests))
- [ X] documentation is changed or added


karellm:
Thanks! Deployed as `7.2.0`

PR #765, opened 2023-01-18
fix: #764 typeArguments may be undefined

### LINKED ISSUE #764: TypeError parsing TypeScript code
## 🐛 Bug Report

Parsing TypeScript code raises a type error when using `TFunction` without type parameters.

Introduced with https://github.com/i18next/i18next-parser/commit/1547997e522579dbcea155d96f8fe8ea3baf2f3a

## To Reproduce

Minimum code example:
```ts
import type { TFunction } from "i18next";
function test(t: TFunction): void { }
```
raises
```
TypeError: Cannot read properties of undefined (reading 'length')
    at JavascriptLexer.functionParamExtractor (node_modules/i18next-parser/dist/lexers/javascript-lexer.js:127:23)
    at parseTree (node_modules/i18next-parser/dist/lexers/javascript-lexer.js:81:41)
    at visitNodes (node_modules/typescript/lib/typescript.js)
```

The following code passes without error:
```ts
import type { TFunction } from "i18next";
function test(t: TFunction<"translation">): void { }
```

Apparently, `node.type.typeArguments` becomes `undefined` if the type does not contain any arguments. This will not be checked in function `functionParamExtractor` (line 127). There is another place in this file that performs this check though (function `expressionExtractor`, line 228).

## Expected behavior

Parser does not exit.

## Your Environment

- *runtime version*: node v18
- *i18next version*: 22.4.9
- *typescript version*: 4.9.4
- *os*: Windows,

## Patch

I will prepare a PR.

danielrentz:

Fixes #764

Fixes #764

### Checklist

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -124,6 +124,7 @@ export default class JavascriptLexer extends BaseLexer {
     ) {
       const { typeArguments } = tFunctionParam.type
       if (
+        typeArguments &&
         typeArguments.length &&
         typeArguments[0].kind === ts.SyntaxKind.LiteralType
       ) {


sibelius:
can we get this merged in? and release a new version ?

src/lexers/javascript-lexer.js
```diff
@@ -124,6 +124,7 @@ export default class JavascriptLexer extends BaseLexer {
     ) {
       const { typeArguments } = tFunctionParam.type
       if (
+        typeArguments &&
```
sibelius, line 127: we also need a fix in the lines below

```jsx
tFunctionParam &&
      tFunctionParam.type &&
      tFunctionParam.type.typeName &&
      tFunctionParam.type.typeName.text === 'TFunction')
```

karellm:
This is deployed as `7.6.0`

src/lexers/javascript-lexer.js
danielrentz: @sibelius Property `typeName` is not optional according to [type definitions](https://github.com/microsoft/TypeScript/blob/main/lib/typescript.d.ts#L946).

sibelius: type definitions can be wrong

Issue #774, opened 2023-02-08
[7.6.0] types not importable


Cielquan:
## 🐛 Bug Report

With #660 (release 7.0.0) a type definition file was added and with #694 (release 7.5.0) the file was added to the `dist` when building. Prior it was missing in the npm package.

I just updated to 7.6.0 and can verify that the types are installed in `node_modules/i18next-parser/dist/index.d.ts`. Unfortunately I cannot import them with `import { UserConfig } from "i18next-parser";` inside `i18next-parser.config.ts`. I get the error that the file cannot be found.

I looked in some other packages installed in my `node_modules` and in most cases I can find a `index.d.ts` file in the package's root directory and not their `dist` directory. When I move the `index.d.ts` file of i18next-parser one directory up in it's root, I can import it with no issues.

I have 0 knowledge about packaging in the JS world, so I cannot comprehend where the real underlying issue is. Maybe I am just too unknowing in this department to correctly import them. But because all those other packages had their types in their root directory I think that's the way to go?

## To Reproduce

1. Install `i18next-parser`
2. create `i18next-parser.config.ts` file
3. add `import { UserConfig } from "i18next-parser";`
4. see error

## Expected behavior

The types should be importable without tinkering.

## Your Environment

- *runtime version*: node v18.10
- *i18next version*: 22.4.9
- *os*: Linux

## Additional Information

I also had the import issue when I build my custom lexer (#678) to support `shouldUnescape`. I could not import the `JsxLexer` class and simply extend it. So I had to fully copy paste the `BaseLexer`, `JavascriptLexer` and `JsxLexer` source to add a couple lines of code.

Like stated above I don't know anything about packaging, but e.g. `i18next-conv` ([repo](https://github.com/i18next/i18next-gettext-converter)) has a cli and I can import it to build custom scripts. They have a file in the `bin/` directory and their source in a `lib/` instead of `dist/` directory and then export everything from a `index.js` file in the package root. Maybe this would be a solution here too?

EDIT: add additional info
EDIT2: add additional info
EDIT3: add additional info

peksipatongeis:
`package.json` is missing the types property and it fails because the types were moved to the dist folder. Created a PR to add the missing types property to fix this.

peksipatongeis:
Meanwhile I fixed the issue by creating a local type declaration `i18next-parser.d.ts`, just make sure it gets included by your tsconfig

```ts
declare module 'i18next-parser' {
  import EventEmitter from 'events';

[...]

PR #775, opened 2023-02-09
Added missing types property to package.json

peksipatongeis:

Added missing types property to package.json to fix issue with types when importing to a .ts file

#774 

### Checklist


karellm:
Thanks, deployed as `7.7.0`

PR #805, opened 2023-04-15
improve warning message when prop variables are detected

josh-egan:

The warning message when a component prop is assigned using a variable currently has a hardcoded value of `Namespace`, but this warning could be fired when any prop is extracted. This PR is to improve the warning message to indicate which prop is using a variable.

No

### Checklist

### DIFF
--- a/test/parser.test.js
+++ b/test/parser.test.js
@@ -2530,7 +2530,23 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
-    it('emits a `warning` event if a namespace contains a variable', (done) => {
+    it('emits a `warning` event if ns prop contains a variable', (done) => {
+      const i18nextParser = new i18nTransform({
+        output: 'test/locales/$LOCALE/$NAMESPACE.json',
+      })
+      const fakeFile = new Vinyl({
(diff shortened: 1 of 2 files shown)


test/parser.test.js
```diff
@@ -2530,7 +2530,23 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
-    it('emits a `warning` event if a namespace contains a variable', (done) => {
+    it('emits a `warning` event if ns prop contains a variable', (done) => {
```
karellm, line 2533: I might be missing something but it looks to me that this test is exactly the same as the following one, isn't it?

karellm (commented):
PR looks good but I've added a comment about the test

test/parser.test.js
ASpwalters: if there's a change requested for this i'll gladly make it, we'd love to have this so we could get back on the main branch of this library =)

karellm: I must have been tired when I reviewed it. Will merge it now

karellm:
Pushed as `8.5.0`

PR #817, opened 2023-05-06
Add support for string concatenation in default values

2Pacalypse-:

When using a string as the second argument to the `t` function, it would be interpreted as `defaultValue` for that key. However, sometimes that string becomes too long and you need to split it into multiple lines. Using template literals has its downsides as it preserves indents and new lines. Thus, the easiest thing to do would be to use string concatenation, e.g.:
```js
i18next.t(
  'key',
  'This is a very long string that needs to be split into multiple ' +
  'lines bla bla bla bla bla bla bla bla bla'
)
```

Not that I'm aware of.

### Checklist

### DIFF
--- a/src/lexers/javascript-lexer.js
+++ b/src/lexers/javascript-lexer.js
@@ -256,13 +256,27 @@ export default class JavascriptLexer extends BaseLexer {
 
       let optionsArgument = node.arguments.shift()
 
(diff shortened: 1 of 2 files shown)


karellm:
Thanks. Deployed as `7.9.0`

PR #824, opened 2023-05-16
Reenable all tests

dhenneke:

I noticed that only a single test is executed since https://github.com/i18next/i18next-parser/commit/10f3d69fc3db47a05197cf784228613418fc4179#diff-fdc851b37e3e94937316ea7b0a3d2cdb4a727931fe4a12772e6ffb8cac29fd53. I assume that this is quite obvious when running the tests for future versions, but better be safe that nothing breaks in the future due to this.

No

### Checklist

### DIFF
--- a/test/parser.test.js
+++ b/test/parser.test.js
@@ -2484,7 +2484,7 @@ describe('parser', () => {
       i18nextParser.end(fakeFile)
     })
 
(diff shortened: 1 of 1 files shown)


karellm:
Thanks!

PR #854, opened 2023-07-11
Improve namespace extraction from hook/HOC

nicegamer7:
While working on localizing our project, we used some code along the lines of the following:
```ts
export default withTranslation(undefined, { withRef: true })(
  ...
);
```

This resulted in the related strings being extracted under the "undefined" namespace, rather than using the default which is expected. Similarly, using a variable causes the strings to be extracted to a namespace that matches the name of that variable, rather than a warning being emitted.

No.

### Checklist


karellm:
Thanks for the thorough PR! It is deployed as `8.2.0`

PR #861, opened 2023-07-19
Extract format parameter from Trans component interpolations

nicegamer7:
While working on localizing our project, we noticed that formatted interpolations in the Trans component weren't being extracted correctly, leading to translations not being loaded correctly in our case.

No.

### Checklist


karellm:
Thanks! Deployed in `8.3.0`

PR #862, opened 2023-07-19
Skip extracting dynamic children in Trans components

nicegamer7:
While working on localizing our project, we found that components with the `i18nIsDynamicList` property weren't being extracted correctly. See https://react.i18next.com/latest/trans-component#using-with-lists-v10.5.0 for the documentation on this prop.

No.

### Checklist


karellm:
Thanks! Deployed in `8.3.0`

Issue #603, opened 2022-06-14
Support TypeScript type casts in JSX


joshkel:
## 🚀 Feature Proposal

i18next-parser cannot currently handle TypeScript `as` casts within a JSX file. For example:

```tsx
<Trans>
  Hello, <strong>{{name} as any}</strong>!
</Trans>
```

gets extracted as `"Hello, <1>{{ name } as any}</1>!"` instead of `"Hello, <1>{{name}}</1>!"`.

## Motivation

As discussed in https://github.com/i18next/react-i18next/issues/1483, React 18's type definitions cause TypeScript to start throwing errors for code such as

```tsx
<Trans>
  Hello, <strong>{{name}}</strong>!
</Trans>
```

since a `{ name }` object is not a valid React child. react-i18next offers a new `allowObjectInHTMLChildren` option to adjust TypeScript's type definitions to allow this; however, it's a fairly drastic option and causes all React children everywhere to allow arbitrary objects (even if they're not part of `<Trans>`).

An alternative is to use an `as any` cast (or similar) at the call site:

```tsx
<Trans>
  Hello, <strong>{{name} as any}</strong>!
</Trans>
```

However, i18next-parser doesn't currently parse this properly.

## Example

See above.

karellm:
Deployed as part of 6.6.0

PR #863, opened 2023-07-20
Allow ignoring typecheck-helper functions around Trans tag variables

pxpeterxu:

Thanks again for maintaining this, @karellm!

This PR helps with extraction for React + Typescript + i18next projects that use a helper function to help with typechecking.

By default, for React with Typescript, components don't allow passing in objects as children. e.g., the below wouldn't typecheck:

```tsx
const SomeComponent = (props) => (
  <Trans>Hello <b>{{ name: props.name }}</b></Trans>
);
```

i18next added [a type option, `allowObjectInHTMLChildren`](https://www.i18next.com/overview/typescript#custom-type-options) that helps with HTML elements. However, this still doesn't work for non-HTML components: e.g., the following would still not pass the typechecker.

```tsx
const SomeComponent = (props) => (
  <Trans>Hello <Link to="/profile">{{ name: props.name }}</Link></Trans>
);
```

As a result, we (and some other teams we know) use either of the below methods:

```tsx
// Option 1: already supported
const SomeComponent = (props) => (
  <Trans>Hello <Link to="/profile">{{ name: props.name } as any}</Link></Trans>
);

// Option 2
const SomeComponent = (props) => (
  <Trans>Hello <Link to="/profile">{{ name: props.name } as unknown as string}</Link></Trans>
);

// Option 3: slightly nicer because we can use a lint rule
// to make sure that castAsString() only appears in <Trans>
const SomeComponent = (props) => (
  <Trans>Hello <Link to="/profile">{castAsString({ name: props.name })}</Link></Trans>
);

function castAsString(obj: Record<string, unknown>): string {
  return obj as unknown as string;
}
```

However, right now, i18next-parser would not pick up `{ name: props.name }` since it's in a function call or multiple levels of type-casts.

### Changes

- Allow passing in an option to the JSX Lexer, `transIdentityFunctionsToIgnore`, that specifies identity functions to "unwrap" before parsing a Trans object for strings
- Allow multiple levels of type-casting (if we're not using an `as any` cast) to be unwrapped

No, but is somewhat related to #603 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,37 @@ If your JSX files have `.js` extension (e.g. create-react-app projects) you shou
 }
 ```
 
+##### Supporting helper functions in Trans tags
+
+If you're working with i18next in Typescript, you might be using a helper function to make sure that objects in <Trans /> components pass the typechecker: e.g.,
+
+```tsx
+const SomeComponent = (props) => (
+  <Trans>
+    Visit
+    <Link to='/user/john'>{castAsString({ name: props.name })}'s profile</Link>
+    {/* Equivalent to, but resolves typechecker errors with */}
+    <Link to='/user/john'>{{ name: props.name }}'s profile</Link>
+  </Trans>
+)
+
+function castAsString(record: Record<string, unknown>): string {
+  return record as unknown as string
+}
+```
+
+In order for the parser to extract variables properly, you can add a list of
+such functions to the lexer options:
+
+```
+{
+  js: [{
+    lexer: 'JsxLexer',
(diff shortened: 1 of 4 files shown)


karellm (commented):
Thanks for the PR, everything looks great but it looks like you removed the README description while I was reviewing the code. I think it would be great if you could put it back so other users have instructions.

pxpeterxu:
Whoops, sorry about that! I realized I'd added some other unneeded code in the README. I just added it back! @karellm take a look when you're free

karellm:
Thanks! Deployed in `8.3.0`

PR #871, opened 2023-08-03
Fix extracting Trans component without key, but with default value

### LINKED ISSUE #249: Trans defaults cause variable names to be treated as keys
## 🐛 Bug Report

The Trans component can be used to interpolate a line with i18next-react templates ("...<1>..."), but not to translate keys (what the `t()` function does). Although I am aware that the Trans component can also perform key translation, there are cases where this is not desired.

In this case, parsing mistakenly extracts a key when nothing should be extracted.
This is a regression that I discovered when upgrading from 3.0.1 to 3.3.0.

## To Reproduce

```js
const line = 'some generated template with <1>bold</1> text';

function Repro() {
  return (
    <Trans
      defaults={line}
      components={[<bold>text</bold>]}
    />
  );
}
```

Parsing should generate an empty file, but produces the following file:
```
{ "line": "line" }
```

## Your Environment

- *runtime version*: i.e. node v13
- *i18next-parser*: 3.3.0
- *i18next*: 19.8.3
- *react-i18next*: 11.7.3
- *os*: Linux

nicegamer7:
Hi again! Just wanted to thank you for maintaining this project, and for responding so quick to my other PRs.

This PR fixes a bug that causes us to extract the keys of Trans components with default values incorrectly.

Take the following example:
```ts
<Trans>
  {{ name }}
</Trans>
```

The above results in the key `{{name}}`, which is correct. However, if we add a default value, it causes us to extract the key incorrectly:
```ts
<Trans defaults="Steve">
  {{ name }}
</Trans>
```

The above results in a key of `Steve`, which isn't correct. The key shouldn't change when we provide a default value.

Possibly fixes #249, but I'm if this is the same issue that the user was describing.

### Checklist

### DIFF
--- a/src/lexers/jsx-lexer.js
+++ b/src/lexers/jsx-lexer.js
@@ -135,10 +135,9 @@ export default class JsxLexer extends JavascriptLexer {
         } else entry[property.name.text] = true
       })
 
+      const nodeAsString = this.nodeToString.call(this, node, sourceText)
       const defaultsProp = getPropValue(tagNode, 'defaults')
-      let defaultValue =
-        defaultsProp || this.nodeToString.call(this, node, sourceText)
-
+      let defaultValue = defaultsProp || nodeAsString
       if (entry.shouldUnescape === true) {
         defaultValue = unescape(defaultValue)
       }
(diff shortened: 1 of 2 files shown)


karellm:
Thanks for the PR, this is deployed as `8.4.0`

nicegamer7:
Thanks karellm.

PR #877, opened 2023-08-08
remove unused dependencies

XaveScor:

I found the unused dependencies in the package.json. This PR removes it.

No

### Checklist


karellm:
Thanks. Pushed as `8.5.0`

PR #881, opened 2023-08-09
Emit warning on non-literal content inside of Trans

nicegamer7:
We noticed in our project that JSX expressions were being extracted incorrectly. We noticed this by setting `debug: true` when initializing i18next, then seeing that expressions were extracted as empty non-self-closing tags.

No.

### Checklist

### DIFF
--- a/test/lexers/jsx-lexer.test.js
+++ b/test/lexers/jsx-lexer.test.js
@@ -233,10 +233,10 @@ describe('JsxLexer', () => {
     it('erases tags from content', (done) => {
       const Lexer = new JsxLexer()
       const content =
-        '<Trans>a<b test={"</b>"}>c<c>z</c></b>{d}<br stuff={y}/></Trans>'
+        '<Trans>a<b test={"</b>"}>c<c>z</c></b><br stuff={y}/></Trans>'
       assert.equal(
         Lexer.extract(content)[0].defaultValue,
-        'a<1>c<1>z</1></1><2></2><3></3>'
+        'a<1>c<1>z</1></1><2></2>'
       )
       done()
     })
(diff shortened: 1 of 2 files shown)


karellm:
@nicegamer7 It looks like a breaking change. The last test looks like it was intended to do what it does. Can you please develop a little more why the current code is broken?

nicegamer7:
@karellm Certainly, my appologies for not including this in the PR initially.

After taking a look at this a bit further, it seems I was incorrect. We can't actually extract expressions correctly, unless they're literal values. I'm going to change this to emit a warning in this case.

Here's the code sandbox where I determined this: https://codesandbox.io/s/long-water-dggns9?file=%2Fsrc%2FApp.tsx

See the debug lines in the console (or the following screenshots) for what the correct extracted value is (the keys that i18next is looking for in the resources).

image
image

image
image

So as seen above, neither key is extracted as the expression text, but rather the value. Since we can't extract the value in this tool unless it's a literal, it makes sense to emit a warning.

karellm:
I'm ok with adding the warning but don't you think that leaving the non-litteral function in the default value provides more context that making it look like it actually worked at extracting it? It is still a breaking change and I want to make sure that we are making the right decision here and not start a ping-pong between both solutions.

nicegamer7:
Okay, I think that makes sense. I revert most of this and just leave the warning.

nicegamer7:
@karellm Just wanted to bump this; take your time if you're busy, no rush.

test/lexers/jsx-lexer.test.js
```diff
@@ -233,10 +233,10 @@ describe('JsxLexer', () => {
     it('erases tags from content', (done) => {
       const Lexer = new JsxLexer()
       const content =
-        '<Trans>a<b test={"</b>"}>c<c>z</c></b>{d}<br stuff={y}/></Trans>'
```
karellm, line 236: Why did you modify theses two tests? Leaving `{d}` in there is more informative don't you think?

nicegamer7: Hmm, I suppose so. Would you like me to assert that a warning is emitted in these tests as well? Or should I just undo the changes?

karellm: I usually prefer tests that check for a single thing as they are easier to maintain but I don't have a strong opinion either way.

nicegamer7: Okay, I tend to agree. I'll leave it out in that case.

karellm:
Thanks, I just published `8.6.0`

nicegamer7:
Thanks karellm.

PR #892, opened 2023-08-22
Fix unescape logic, expression props, and ICU format for Trans component

### LINKED ISSUE #886: Keys extracted with escaped characters starting in v8.4.0
## 🐛 Bug Report

Up to v8.3.0, the following code would result in the key being extracted as `"Don't click me"`:

```js
<Trans shouldUnescape>Don&apos;t click me</Trans>
```

Starting in v8.4.0, it gets extracted as `"Don&apos;t click me"`. The main reason this is an issue is that i18next will look for the key as "Don't click me", disregarding of whether `shouldUnescape` is `true` or `false` (I reported that as a separate issue [here](https://github.com/i18next/react-i18next/issues/1662)).

This basically means that all the keys extracted by i18next-parser which contained escaped characters will be ignored by i18next.

## Your Environment

- *runtime version*: node v18
- *i18next version*: 23.4.4
- *react-i18next*: 13.1.2
- *os*: MacOs

nicegamer7:
After #871, it seems I introduced a regression, though it did help find a bug.

~~It seems that react-i18next does indeed default the key to the `defaults` prop (I added a comment to prevent anyone from making the same mistake again). I've reverted the changes from #871 to match react-i18next's behavior.~~ After reporting this on react-i18next's bug tracker (i18next/react-i18next#1663), it looks like this was unintended. I fixed that, so no need to revert #871.

Otherwise, it seems we had a bug where we were unescaping values in reverse (i.e. unescaping when it wasn't needed, and not unescaping when it was needed). I corrected this, and also made it so that we always unescape the key, since this matches react-i18next's behavior as well.

Closes #886.

### Checklist

### DIFF
--- a/test/lexers/jsx-lexer.test.js
+++ b/test/lexers/jsx-lexer.test.js
@@ -452,12 +452,20 @@ describe('JsxLexer', () => {
         done()
       })
 
+      it('unescapes key when i18nKey is not provided', (done) => {
+        const Lexer = new JsxLexer()
+        const content = '<Trans>I&apos;m Cielquan</Trans>'
+        assert.equal(Lexer.extract(content)[0].key, "I'm Cielquan")
+        done()
+      })
+
       it('supports the shouldUnescape options', (done) => {
         const Lexer = new JsxLexer()
-        const content = '<Trans shouldUnescape>I&apros;m Cielquan</Trans>'
+        const content = '<Trans shouldUnescape>I&apos;m Cielquan</Trans>'
+        assert.equal(Lexer.extract(content)[0].key, "I'm Cielquan")
         assert.equal(
           Lexer.extract(content)[0].defaultValue,
-          'I&apros;m Cielquan'
+          'I&apos;m Cielquan'
         )
         done()
       })
--- a/test/lexers/jsx-lexer.test.js
+++ b/test/lexers/jsx-lexer.test.js
@@ -132,6 +132,29 @@ describe('JsxLexer', () => {
       done()
     })
 
+    it('extracts keys from Trans elements without an i18nKey, with defaults, and without children', (done) => {
+      const Lexer = new JsxLexer()
+      // Based on https://react.i18next.com/latest/trans-component#alternative-usage-components-array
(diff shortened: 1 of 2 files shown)


nicegamer7:
Actually, I'm going to put this on hold until i18next/react-i18next#1663 is resolved, since it might mean we don't have to revert #871.

test/lexers/jsx-lexer.test.js
```diff
@@ -452,12 +452,20 @@ describe('JsxLexer', () => {
...
       })
 
+      it('unescapes key when i18nKey is not provided', (done) => {
+        const Lexer = new JsxLexer()
+        const content = '<Trans>I&apos;m Cielquan</Trans>'
+        assert.equal(Lexer.extract(content)[0].key, "I'm Cielquan")
+        done()
+      })
+
       it('supports the shouldUnescape options', (done) => {
```
karellm, line 462: I find this a little counter intuitive and confusing. I would expect that `shouldUnescape` would unescape the `defaultValue`. If I understand correctly, that's not the case in `react-18next`, is that right?

Also, I'm wondering if we should add a test with a `i18nKey` property with escaped value. wdyt?

nicegamer7: It is a little confusing, but I'll try to explain as I understand it.

The `shouldUnescape` option means that after the translation is loaded/displayed, it will be unescaped. That's why when we see the `shouldUnescape` option, we **don't** unescape when extracting, so that it can be unescaped by react-i18next when it is loaded/displayed.

Also, I think it makes sense to add a test for a `i18nKey` with an escaped value, nice suggestion.

nicegamer7:
Hey @karellm, I just pushed another commit that fixes two more issues that I noticed. The two issues are:
1. Extracting Trans components with the ICU format was broken
2. Extracting props with object/array expressions was broken

Take a look when you have a chance and let me know if there are any concerns!

karellm:
LGTM! Thanks for the great work and it is now deployed as `8.7.0`

Issue #912, opened 2023-09-14
Support custom namespaced functions and components


taneliang:
## 🐛 Bug Report

In our codebase, we'd like to re-export i18n functions from a common namespace, i.e. instead of using `t`, `i18next.t`, or `<Trans>`, we'd like to use something like `Strings.get` and `<Strings.Text>`. Although the JS and JSX lexers support custom `functions` and `componentFunctions`, they assume that the functions and components are TS identifiers, and are not able to match the names of property accessors.

## To Reproduce

Given this config:

```js
module.exports = {
  lexers: {
    js: [
      {
        lexer: 'JavascriptLexer',
        functions: ['strings.for'], // Setting this to `'for'` works, but this will also match other functions named `for`, including `Symbol.for`.
      },
    ],
    jsx: [
      {
        lexer: 'JsxLexer',
        functions: ['strings.for'],
        componentFunctions: ['strings.Text'],
      },
    ],
  },
}
```

And this code:

```jsx
Strings.get('key1');
<Strings.text i18nKey="key2" />
```

## Expected behavior

Both keys should be matched and extracted.

```json
{
  "key1": "",
  "key2": ""
}
```

## Your Environment

- *runtime version*: Node
- *i18next version*: 23.5.1
- *os*: Mac

karellm:
Fixed as of `8.8.0`

PR #913, opened 2023-09-14
Support custom namespaced functions and components

taneliang:

To support matching namespaced functions and components like `Strings.get` and `<Strings.Text>`.

Yes https://github.com/i18next/i18next-parser/issues/912

### Checklist

- [ ] documentation is changed or added – I'll be happy to add docs for this, but the existing docs should already be enough


karellm:
Thanks, this is in `8.8.0`

PR #919, opened 2023-09-28
Update index.d.ts

Cielquan:

I removed two config values from the UserConfig type which were removed back in v7.0.
I also updated the defaultValue function type to include the 'value' argument which was missing.

No

### Checklist


karellm:
Thanks, this is in `8.8.0`

Cielquan:
Thanks for merge and fast release

PR #924, opened 2023-10-02
fix c&p error

janpaepke:

just a small copy/paste mistake

No

### Checklist


karellm:
Thanks!

Issue #684, opened 2022-11-09
High vulnerability in dependency package vinyl-fs > glob-stream > glob-parent


pedrodlz:
## 🐛 Bug Report

High vulnerability found in package glob-parent
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high                               │ glob-parent before 5.1.2 vulnerable to Regular Expression                                                                   │
│                                       │ Denial of Service in enclosure regex                                                                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package                         │ glob-parent                                                                                                                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in                      │ >=5.1.2                                                                                                                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of              │ i18next-parser                                                                                                                                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path                               │ i18next-parser > vinyl-fs > glob-stream > glob-parent                                                                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info                       │ https://www.npmjs.com/advisories/1081884                                                                                        │
└───────────────┴──────────────────────────────────────────────────────────────┘

## To Reproduce

```js
yarn audit
```

## Expected behavior

No vulnerabilities found

## Your Environment

- *runtime version*: node v16.16.0, Ract Native 0.68.5
- *i18next version*: 6.6.0
- *os*: Windows

eelkevdbos:
If you run npm v8.3, you can add an override to package.json until it is fixed:

```json
  "overrides": {
    "glob-parent": "^5.1.2"
  }
```

chidg:
Hi maintainers, any chance of an update on this? `overrides/resolutions` is an ok workaround, but it would be nicer if there was a proper fix.

karellm:
@chidg I have limited bandwidth atm, would you be available to make a PR?

chidg:
Hi @karellm, it turns out you already have a PR for the dependency upgrade: #859 
Unfortunately there is a failing test relating to Broccoli. I spent some time debugging it but couldn't make much progress.

karellm:
I've spent about an hour trying to debug this and couldn't find why the test fail. It looks like the files are not being parsed for some obscure reason.

chidg:
Same experience I had!

woutervanvliet:
[...]

PR #930, opened 2023-10-13
Upgrade vinyl-fs from 3.0.2 to 4.0.0

woutervanvliet:

There is a high vulnerability attached to vinyl-fs' dependency tree, blocking me from being able to use this package in my own project. The automatic dependabot PR can't merge, due to changes in glob-handling, which I'm also handling in this PR.

Yes #684 

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,9 @@ i18n = new i18nextParser([i18n], {
 export default i18n
 ```
 
+> **Note**: You may need to configure Broccoli to place temporary files (option: tmpdir) within the current working
+> directory as I18next-parser does not traverse down beyond that.
+
 ## Options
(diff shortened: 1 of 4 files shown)


woutervanvliet:
Turned out glob-stream, used by vinyl-fs, only scans from the current directory and upwards, thus not finding temporary files created by Broccoli and in turn not creating translation files.

I don't have experience with Broccoli though, so I'm a bit unsure if this change will have any implications on projects using i18next-parser with Broccoli. Added a note though to the readme file that hopefully provides enough information for Broccoli users.

karellm:
Thanks! This is deployed as `8.9.0`

PR #932, opened 2023-10-14
Sort & failOnUpdate

### LINKED ISSUE #926: Changes in order should cause --fail-on-update to fail
## 🐛 Bug Report

When the keys are complete but in a wrong order calling `i18next --fail-on-update` with `sort: true` in the config does update the files but not fail.

## To Reproduce

1. Have a project with an unsorted translation file and `sort: true` in the config.
2. Run `i18next --fail-on-update 'src/**/*.{ts,tsx,js,jsx}'`

## Expected behavior

Expected the command to fail, since there are updates in the translation file.

## Your Environment

- *runtime version*: node v20
- *i18next-parser version*: 8.3.0
- *os*: Windows

mspiess:

Exit with exit code 1 when `sort` and `failOnUpdate` are set and keys are not already sorted

Fixes #926

### Checklist

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -296,7 +296,20 @@ export default class i18nTransform extends Transform {
             typeof sort === 'function'
               ? sort
               : makeDefaultSort(this.options.pluralSeparator)
-          maybeSortedNewCatalog = sortKeys(newCatalog, { deep: true, compare })
+          let wrappedCompare = this.options.failOnUpdate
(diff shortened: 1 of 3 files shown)


mspiess:
@karellm any chance to get some 👀 on this?

karellm:
Thanks for the PR! Will be deployed as `8.11.0`

Issue #666, opened 2022-10-27
`keepRemoved` option makes the `failOnUpdate` fails without any update


ThomasF34:
## 🐛 Bug Report

When using the `keepRemoved` option, set to true, the `failOnUpdate` is failing though no update is made 

## To Reproduce

- Use `keepRemoved: true` option
- Have some unreferenced keys 
- Run the cli with `--fail-on-update`
- The command fails though no update is made

## Expected behavior

As unreferenced keys are expected when you use the `keepRemoved: true` option, I would expect the `fail-to-update` run to exit without error

## Note
(having a quick look at [this code](https://github.com/i18next/i18next-parser/blob/42b66c1dca5986f462825057e10db58222f5ddc0/src/transform.js#L267-L284) it seems the `oldCount` is taken into account even with the `keepRemove` option)

## Your Environment

- *runtime version*: node v16.13.0
- *i18next version*: tested on 6.5.0 and 6.6.0
- *os*: Mac

tran-simon:
I have the same issue

karellm:
Can one of you make a PR?

sabbirzzaman:
@ThomasF34 Is there any option to disable `i18next-parser` from removing my unused keys? i18n-parser is just removing all my dynamic keys. And this is not appropriate for me totally. I just want to keep the empty key generating feature for other languages. Is it possible?

leandro-gong:
Any updates on this issue?
I'm also need to use `keepRemoved = false` and `failOnUpdate = true` but it's failing with no changes

cravend:
PR created! @karellm

Twister42:
Just made a new PR with the tests @karellm

karellm:
Thanks @Twister42  Should be deployed shorlty as `8.10.0`

PR #948, opened 2023-12-18
fix: keepRemoved option makes the failOnUpdate fails without any update

Twister42:

Fixes the issue 666

Yes/No #666


karellm:
Thanks! Will make it to `8.10.0`

PR #957, opened 2024-01-02
Fix --fail-on-update false negatives from new sort code.

### LINKED ISSUE #955: Regression with version 8.0.11
## 🐛 Bug Report

After upgrade to 8.0.11, this command fail:
```
i18next 'src/**/*.{ts,tsx}' --config i18next-parser.config.js --fail-on-update --silent
```

With

```
> i18next 'src/**/*.{ts,tsx}' --config i18next-parser.config.js --fail-on-update --silent

  [error]   Some keys were sorted and failOnUpdate option is enabled. Exiting...
npm ERR! Lifecycle script `i18n:check` failed with error: 
npm ERR! Error: command failed
```

Sort is enabled.

If I remove `--fail-on-update`, no file are updated, so no need to sort anything.

## To Reproduce

Run the command above with this config file:

```javascript

module.exports = {
  contextSeparator: '_',
  createOldCatalogs: false,
  defaultValue: '',
  indentation: 2,
  keepRemoved: false,
  keySeparator: '.',
  lexers: {
    hbs: ['HandlebarsLexer'],

    htm: ['HTMLLexer'],
    html: ['HTMLLexer'],

    mjs: ['JavascriptLexer'],
    ts: ['JsxLexer'],
    jsx: ['JsxLexer'],
    tsx: ['JsxLexer'],

    default: ['JavascriptLexer'],
  },
  lineEnding: 'lf',
  locales: ['en', 'fr'],
  namespaceSeparator: ':',
  output: 'src/locales/$LOCALE/$NAMESPACE.json',
  pluralSeparator: '_',
  input: undefined,
  sort: true,
  skipDefaultValues: false,
  useKeysAsDefaultValue: false,
  verbose: false,
  failOnWarnings: false,
  failOnUpdate: false,
  customValueTemplate: null,
  resetDefaultValueLocale: null,
  i18nextOptions: null,
};
```

## Expected behavior

No file need to be sorted so I should not have any error.

## Your Environment

- *runtime version*: node v21.5.0
- *i18next version*: 23.7.12
- *i18next-parser version*: 8.11.0
- *os*: Mac

tec27:
This replaces the code that previously tried to check for sorting via a wrapped `compare` function with a much simpler approach: JSON encode the old and (sorted) new version and see if they match. This code only runs if we have not already seen an update to the keys/contents before, so the performance implications are minimal.

This also adds a test to ensure that false negatives don't occur (i.e. it checks the --fail-on-update behavior against a file that is not being updated and was already sorted).

Fixes #955.

The previous update broke --fail-on-update, making it always fail if sorting was configured.

Yes #955

### Checklist

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -298,21 +298,21 @@ export default class i18nTransform extends Transform {
             typeof sort === 'function'
               ? sort
               : makeDefaultSort(this.options.pluralSeparator)
-          let wrappedCompare = this.options.failOnUpdate
-            ? (...args) => {
-                const result = compare(...args)
-                if (result > 0) {
-                  this.parserHadSortUpdate = true
-                  wrappedCompare = compare
-                }
-                return result
-              }
-            : compare
           maybeSortedNewCatalog = sortKeys(newCatalog, {
             deep: true,
-            compare: wrappedCompare,
+            compare,
           })
+
           maybeSortedOldCatalog = sortKeys(oldCatalog, { deep: true, compare })
+
+          if (this.options.failOnUpdate && !this.parserHadUpdate) {
+            // No updates to the catalog, so we do the deeper check to ensure it is also sorted.
+            // This is easiest to accomplish by simply converting both objects to JSON, as we'd have
+            // to do a deep comparison anyway
+            this.parserHadSortUpdate =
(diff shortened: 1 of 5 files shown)


karellm:
Thanks for the PR, this is deployed as `8.12.0`

PR #970, opened 2024-02-01
Output the files in which a string is found in customValueTemplate

pxpeterxu:

Thanks again for the great framework and maintenance @karellm -- we've now translated our [Wanderlog](https://wanderlog.com) trip planner mobile app to French with help from this!

As a part of the process, we realized that sometimes, translators can benefit with some extra context about the strings they're translating: e.g., it matters if a string like "Round trip" comes from a file called FlightSearchInput.tsx. The `customValueTemplates` have been really helpful with other annotations, so we thought it'd be helpful to extend it so that we can put:

```ts
{
  locales: [SupportedLocale.en],
  output: 'forTranslators.json',
  customValueTemplate: {
    message: '${defaultValue}',
    filePaths: '${filePaths}',
  },
}
```

### Changes

- Add an extra special-case for the customValueTemplate handling (similar to `defaultValue`) which will also substitute in the paths of all the files where we found a particular key
- Add tests to verify this behavior works properly with multiple files

**Note**: to make this easier to merge, I also invited you to have write access to our repository.

Example output when run on our codebase:

```json
{
  "Perfect for road trips and saving $$$ on gas! Get the best route auto-rearranged.": {
    "filePaths": [
      "/Users/peter/Code/itineraries/client/templates/LandingPage/LandingPageProFeatureTiles.tsx"
    ]
  },
  "See more": {
    "filePaths": [
      "/Users/peter/Code/itineraries/client/templates/LandingPage/LandingPageReview.tsx",
      "/Users/peter/Code/itineraries/client/templates/LandingPage/LandingPageTestimonial.tsx"
    ]
  }
}
```

No

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,13 +204,33 @@ export default {
 
   customValueTemplate: null,
   // If you wish to customize the value output the value as an object, you can set your own format.
-  // ${defaultValue} is the default value you set in your translation function.
-  // Any other custom property will be automatically extracted.
+  //
+  // - ${defaultValue} is the default value you set in your translation function.
+  // - ${filePaths} will be expanded to an array that contains the absolute
+  //   file paths where the translations originated in, in case e.g., you need
+  //   to provide translators with context
+  //
+  // Any other custom property will be automatically extracted from the 2nd
+  // argument of your `t()` function or tOptions in <Trans tOptions={...} />
   //
   // Example:
-  // {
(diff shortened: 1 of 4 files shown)


karellm:
Thanks @pxpeterxu ! This seems like a valuable addition. I will make it to `8.13.0`

PR #981, opened 2024-02-26
Fix example in README.md

obiwan00:
`HtmlLexer` in config leads to build error. The correct parameter is `HTMLLexer`


karellm:
Thanks!

Issue #617, opened 2022-07-19
Vue 3 support


leokolezhuk:
## 🚀 Feature Proposal

Currently the usage of vue-template-compiler in this package prevents using the parser in a vue 3 project.

Supporting both vue 2 and vue 3 versions would be very beneficial I believe.
## Example

Please provide an example for how this feature would be used.

Toilal:
Maybe a simple solution would be to make `vue-template-compiler` a peerDependency.

leokolezhuk:
> You are welcome to make a PR to solve the situation! I will review and merge it.

Hi @karellm,

I am open to doing that.
But while thinking about how to approach it best, I am coming to the conclusion that this **VueLexer might not even be needed at all**. Please, correct me if I am wrong on this.

So in Vue SFC files you have the template and script sections that can contain localizable messages. Regardless of where the message is defined, template or script, it will still be wrapped in a function $t() or t(). Therefore, it seems to me that the **JavaScriptLexer alone is enough to deal with this case**. That is how I am using the parser in my project as well - with the same lexer config for js and vue files. There is no need to compile the .vue file as the VueLexer does it, you can just parse it as a normal JS file then. 

The only case that is not covered with this approach is when one would use a i18next vue wrapper like [this one ](https://panter.github.io/vue-i18next/guide/started.html) and uses the directives(HTML attributes) to define localizable messages. But this case is not covered with the current VueLexer either. But if you use the HTML lexer in combination with the JS lexer for a case like this, you should be covered. 

To sum up at this point I actually see no reason for the VueLexer to exist even for vue users. 
Removing it would also remove the nasty vue-template-compiler dependency that prevents usage with vue3. 

What do you think, @karellm ?

karellm:
Should be fixed as of `9.0.0`

poicooler:
I seem to have a slight issue using version 9.0.0 with vue 3.

When using the translation function in a prop, it is not detected by the parser.

Exemple in some vue files:
```
<template>
  <button :aria-label="t('button aria label')">
    {{ t('button label') }}
  </button>
</template>

...
```

In this case, "button label" is detected and thats fine. However, "button aria label" is not detected and is ignored.

daniel100097:
Does anyone have a solution for parsing attribute translations ? 

like this:

:attr="t('languagekey')"

PBK-B:
[...]

PR #1007, opened 2024-05-28
Remove VueLexer

jolson474:

While investigating how to run i18next-parser on a vue3 code base, I came across this ticket which suggested simply removing `VueLexer`, which would also eliminate the dependency issues with `vue-template`:
https://github.com/i18next/i18next-parser/issues/617

After trying it, this change seemed to meet my needs very nicely, so I thought I would contribute it back to make things easier for other folks using Vue3 (and dropping support for Vue2 didn't seem like it would be a very surprising change going forward).

In addition to removing VueLexer from the code (and its tests) I also updated the documentation and bumped the version to 9.0 since this is a breaking change.  The migration documentation is updated to reflect that you need to remove `VueLexer` from any configuration files as well.

Let me know if you have any concerns/questions with this approach.  Thanks!

Yes #617

### Checklist


karellm:

PR #1008, opened 2024-05-29
custom contextSeparator fix

vseva:

Unused keys where words separated with underscore (for example "some_key", "some_key_just_started_like_previous") not removed because _ hardcoded in helpers.js contextRegex and both "some_key" and "some_key_just_started_like_previous" treats like context forms of one key.

No

### Checklist


vseva:
@karellm please look at this PR

karellm:

Issue #1015, opened 2024-06-20
Plural keys not detected anymore when `count` is an expression


cochery:
## 💥 Regression Report

`t('mykey', '{{count}} something', { count: expression() })` used to be extracted as 2 plural keys, but the plural form isn't detected anymore.

`t('mykey', '{{count}} something', { count: value })` is still extracted as 2 plural keys.

## Last working version

Worked up to version: 8.13.0

Stopped working in version: 9.0.0

## To Reproduce

```js
function expression() {
  return 1;
}
`t('mykey', '{{count}} something', { count: expression() })`
```

Only one key `mykey` is extracted.

## Expected behavior

2 keys `mykey_one` and `mykey_other` should be extracted in English. That was the former behaviour and still works when count is a value and not an expression.

Probably introduced by https://github.com/i18next/i18next-parser/pull/994

## Your Environment

- *runtime version*: node v20.10.0
- *i18next version*: i.e. 23.11.5
- *os*: Mac

glsignal:
Ran into this issue as well. I opened a PR with a patch here -> https://github.com/i18next/i18next-parser/pull/1022

karellm:
This is now part of `9.0.1`.

PR #1022, opened 2024-07-15
Fix plurals not being detected when count is a CallExpression

glsignal:

Addresses https://github.com/i18next/i18next-parser/issues/1015 where the JavaScript lexer would no longer correctly identify plurals where the count was a CallExpression

Yes #1015

### Checklist

- [x] ~documentation is changed or added~


karellm:
This is now part of `9.0.1`. Thanks for the PR!

Issue #1038, opened 2024-08-01
React component key is extracted instead of i18nKey


Shiwa:
## 🐛 Bug Report

When extracting translation from a `Trans` component with a react `key` prop, this key is used in place of the `i18nKey`.

## To Reproduce

Running i18next-parser on the following tsx code

```js
<SomeComponent
  texts={[
    <Trans ns="settings" key={0} i18nKey="my.i18n.key">
      <span className="font-bold">Lorem Ipsum…</span> 
    </Trans>,
  ]}
/>
```

## Expected behavior

The etracted key should be `my.i18n.key` but is instead `{0}`

```js
{
  "{0}": "<0>Êtes-vous sûr de vouloir déconnecter votre compte ?</0> En déconnectant votre compte Creditsafe, le portefolio associé à votre compte LeanPay sera alors supprimé ainsi que la surveillance des tiers qui y sont liés."
}
```

## Your Environment

- *runtime version*: node 20, pnpm
- *i18next version*: 23.6.0 ;   react-i18next: 13.3.1 ;  i18next-parser: 8.13.0
- *os*: Linux
- I tried quickly with i18next-parser and by customizing the JsxLexer, but got the same result. Changing the prop order of key / i18nKey did not fix the problem either, key is always used in place of i18nKey

adrai:
If you like you could try if it works with: https://github.com/i18next/i18next-cli

PR #1042, opened 2024-08-05
Fix JSX key attribute clobbering detected value

comp615:

React components have a protected attribute "key" which can seemingly clobber the detected/parsed i18n key due to a naming conflict.

Yes #1038

### Checklist


karellm:
@comp615 I have mixed feeling about this as some people might be relying on the `key` attribute. I'd prefer a fix that ignores the `key` attribute only if the `i18nKey` one is present. Can you update the PR?

PR #1046, opened 2024-08-10
chore: repo maintenance

ripecosta:

I noticed some update and maintenance PRs on the repo and decided to lend a hand. Here's a list of what this PR contains:

- Updated all dependencies to latest version to force downstream updates. Most libraries were already up to date in the yarn.lock. This should prevent any issues where `i18next-parser` uses a feature from a dependency that's only available in the latest version (local yarn.lock) but not in the lowest version is declares (package.json). Makes it easier to understand which version the package is using without having to dig through the lockfile too.
   - Two exception here: `typescript` and `i18next` were not updated to avoid potential dependency mismatch issues downstream. TS especially has a tendency to cause these.
   - `chai` got a major version upgrade to v5 ([changelog](https://github.com/chaijs/chai/releases/tag/v5.0.0)) - no breaking changes relevant to this project.
- Deduplicated yarn.lock dependencies. Not that the repo was slow to install but it should be even faster now.

No.

### Checklist


.husky/.gitignore
ripecosta, line 1: Husky v9 no longer needs this .gitignore file. The `_` folder itself contains one that does exactly the same.

.husky/pre-commit
ripecosta, line 1: Husky v9 no longer requires this file to be have executable permissions.

.github/dependabot.yml
```diff
@@ -5,8 +5,10 @@ updates:
   target-branch: master
   schedule:
     interval: daily
+  versioning-strategy: increase-if-necessary
```
ripecosta, line 8: Let me know what your thoughts are regarding using `increase` here instead.

ripecosta:
@karellm I raised this PR with some maintenance for the repo if you could have a quick look. Let me know if you anything isn't clear or if you'd like me to split it into smaller PRs.

I was also thinking about raising a separate PR to improve the linting config, if that ok? Nothing major, just quality of life stuff.

.github/dependabot.yml
karellm: @ripecosta I'm not the one who set this up in the first place and I'm not familiar with the options, can you make a case as to why one is better than the other?

karellm:
@ripecosta Thanks for the PR. 

On a side note, please don't do dependency updates alongside code improvements, I prefer multiple PRs with a single focus for each PR as it makes it easier to merge.

.github/dependabot.yml
ripecosta: Sure thing. I'll also break the PR up into 3 separate ones so it's easier to review/merge.

ripecosta: Updated the PR. All it does now is update dependencies. The other bits I put into:
- #1053 
- #1054 
- #1055

PR #1053, opened 2024-08-12
fix: node engine versions

ripecosta:

Fixes node engine version to match [the version documentation](https://github.com/i18next/i18next-parser/blob/master/README.md#versions)

_Note_: `>=18.0.0 || >=20.0.0 || >=22.0.0` is the same as `>=18.0.0` so I assumed the desired behaviour is `^18.0.0 || ^20.0.0 || ^22.0.0` in accordance to the docs. Please let me know if that's not the case.

No

### Checklist


package.json
```diff
@@ -11,7 +11,7 @@
     "i18next": "./bin/cli.js"
   },
   "engines": {
-    "node": ">=18.0.0 || >=20.0.0 || >=22.0.0",
+    "node": "^18.0.0 || ^20.0.0 || ^22.0.0",
```
ripecosta, line 14: This change will match the docs but it will make it impossible to use the lib with node v19 or v21 without passing a flag to ignore engines. If there's undesirable behaviour using `>=18.0.0 <22.0.0 || ^22.0.0` is better.

antonmedv:
Why these restrictions needed? Now node23 is out, but we can't install it.

PR #1054, opened 2024-08-12
chore: upgrade husky config to v9

ripecosta:

Upgrade husky config to v9. V9 was already in use but still using the legacy config method.
Also replaces the npx commands with yarn commands because there's not need to use npx since yarn already installs lint-staged.

No

### Checklist


.husky/.gitignore
ripecosta, line 1: Husky v9 no longer needs this `.gitignore` file. The `_` folder itself contains one that does exactly the same.

ripecosta:
Rebased the PR to fix merge conflicts and it's ready to go now.

CC: @karellm

karellm:
You bit to it, thanks!

PR #1057, opened 2024-08-15
Add feature to skip identical keys for specified locales

### LINKED ISSUE #1036: Config to remove key/value, if key=value
## 🚀 Feature Proposal

A config to not add or remove a key/value from the translation files.

1. First, there is a need to define a default language.
2. Then in default, if `key = value` or `value = ""`, then it can be removed from the translation file.

Basically the first point is not _required_, but I think it makes sense to limit users to have a structure to always have keys in one language, not different languages!

## Motivation

In some cases, it doesn't make sense to have a diff key, and you have the "value" as the "key" (like gettext)
Some apps are loaded mostly with a default language, so there is no need to load a language file at all.

Aryan-mor:
Implemented logic to skip keys with identical values for specified locales in the i18next parser. Fixes #1036 

Please refer to #1036

Yes #1036

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+# X.X.X
+
+- Add skipIdenticals option #1036
+
 # 9.0.2
 
 - Fix cheerio dependency #1045
--- a/README.md
+++ b/README.md
@@ -253,6 +253,9 @@ export default {
   // {
   //   lineWidth: -1,
   // }
+
+  skipIdenticals: [],
+  // An array of locales to skip adding entries which their keys are identical to their values or their values are empty string.
 }
 ```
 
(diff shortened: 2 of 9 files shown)


karellm:
I'm a bit puzzled by this use case but I don't really like the implementation. 

First it looks very specific to a use case you have (not something many people would need). Second the option name `skipIdenticals` shouldn't remove empty values imo. That said, I don't want to make the API too complex with an extra `skipEmtpy`.

I actually think that the best solution here is to add hooks. We could have an hook to update the catalog before it is written to file. You could then add the logic you need in there.

What do you think?

omid:
@karellm I can answer about the use case. If you have worked with gettext (i18next has gettext tag on npm) you know in gettext, you write translations like `t('The Default Language Text')` and mostly the default language (which is EN by default) translation file is almost empty.

We can rename the name `skipIdentical` to elaborate the use case.
I agree, and we can have another array for `skipEmpty` to skip empties in _some_ languages.
On the other hand, it can be more objective, like refer to `defaultLanguages`.

PR #1058, opened 2024-08-19
fix: invalid dependabot config

ripecosta:

The `versioning-strategy` was set to `increase` for both npm and github actions in #1055. However, the GitHub Actions package ecosystem does not support that options ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy)), which caused [this validation error](https://github.com/i18next/i18next-parser/runs/28938435374) when the change was merged.
This PR fixes that error by removing the `versioning-strategy` option from the GH Actions package ecosystem settings.

No.

### Checklist


ripecosta:
@karellm Fix for an issue introduced by one of my earlier PRs.

Oersta1rdt:
Good

Issue #973, opened 2024-02-19
Incorrect extraction for multiple `useTranslation` on the same page


affanshahid:
## 🐛 Bug Report

If I have multiple invocations of `useTranslation` using different namespaces on the same page, the extraction does not work correctly.

## To Reproduce
Create a component like:

```tsx
const useSchema = () => {
  const { t } = useTranslation('validation');
  t('errors.invalid_type_received_undefined') // Note: `errors.invalid_type_received_undefined` is in ns validation
  // ...
};



export const HomePlanStep = () => {
  const { t } = useTranslation(['application-address', 'common']);
  useSchema()

  return <p>{t('homePlanStep.submitBtnText')}</p> // Note: this is in ns application-address
};
```

The dictionary that gets generated incorrectly has `errors.invalid_type_received_undefined` associated with NS: 'application-address'. The actual app runs just fine like this, only the extraction generates incorrect dictionaries.

If I make the following change, the dictionary gets generated correctly:
```tsx
const useSchema = () => {
  const { t } = useTranslation('validation');
  t('errors.invalid_type_received_undefined', { ns: 'validation' }) // Note: added explicit ns
  // ...
};
```
Now the dictionary does not associate `errors.invalid_type_received_undefined` with NS: 'application-address'.

## Expected behavior

The extractor should not mix-up namespaces when extracting

## Your Environment

- *runtime version*: Node v18.19.0
- *i18next version*: v23.7.16
- *i18next-parser version*: v8.12.0
- *os*: MacOS Sonoma 14.2.1

karellm:
Should be fixed with #1083. Deployed as `9.3.0`

Pouet--:
It still does not work for different keyPrefix values. I made a post on the PR [here](https://github.com/i18next/i18next-parser/pull/1083#issuecomment-2685101123)

rocky-luo-opus:
It doesn't work, see [1137](https://github.com/i18next/i18next-parser/issues/1137)

Issue #494, opened 2021-12-22
Key extraction to wrong namespace


tylerrasor:
## 🐛 Bug Report

When using the `useTranslation` hook to get two different TFunctions scoped to different namespaces, extraction should place keys into the namespace of the corresponding TFunction.

## To Reproduce

```js
const { t } = useTranslation();
const { t: otherT } = useTranslations('otherNs');
const header = t('header string');
const otherHeader = otherT('other string that should not be extracted');
```

## Expected behavior

Should get extracted to the default namespace:
```js
// translation.json
"header string": "header string"
```

But gets extracted to the namespace provided to the other `useTranslation` hook:
```js
// otherNs.json
"header string": "header string"
```

____

I believe this is the corresponding test that should be added, but haven't gotten much farther than that
```js
// javascript-lexer.test.js
it('uses default namespace when other t function', () => {
    const Lexer = new JavascriptLexer()
    const content =
      'const {t} = useTranslation("baz");' +
      'const {t:otherT} = useTranslation("otherNs");' +
      't("bar");'
    assert.deepEqual(Lexer.extract(content), [
      { namespace: 'baz', key: 'bar', ns: 'baz' },
    ])
  })
```

![image](https://user-images.githubusercontent.com/4198251/147149311-f3916725-a635-4b66-96fa-233b71bf4392.png)

## Your Environment

- *runtime version*: node v14
- *i18next version*: 21.5.2
- *i18next-parser version*: 5.3.0
- *os*: Mac
- *any other relevant information*

karellm:
@tylerrasor It is indeed a missing feature at the moment.

tux21b:
I just discovered the same bug with key prefixes instead of namespaces. Looks like i18next-parser threads all t functions with the config that appeared last.

```
const { t: tRoot } = useTranslation();
const { t } = useTranslation('translation', {keyPrefix: 'staticData.locations'});
```

cah4a:
any progress here?

Bladieblah:
Bump

thomasmery:
Hi there,

Got bit by this again today ...

unfortunately can't volunteer for a PR

but would really love to see this solved

thanks

karellm:
Should be fixed with #1083. Deployed as `9.3.0`

Pouet--:
@karellm Thanks a lot ! It there a way to upgrade to this version right now ?

Pouet--:
It still does not work for keyPrefix with version 9.3.0.
[...]

PR #1083, opened 2024-10-26
[Fix] Allow to use multiple translation functions with different namespaces and keyPrefixes

### LINKED ISSUE #737: The keyPrefix argument is ignored in JavascriptLexer, and not in JsxLexer
## To Reproduce

Name this file with .ts extension (configured with JavascriptLexer) and the keyPrefix is ignored, if you use JsxLexer instead, it's works
```typescript
export default function useArticlesFilter() {
  const { t } = useTranslation('filters', { keyPrefix: 'articles' });

  return useMemo(() => {
    return {
      menuTitle: t('menu'),
      badge: {
        Component: ArticlesBadge,
      },
      panel: {
        Component: ArticlesPanel,
        query: ArticlesPanelQuery,
      },
    };
  }, [t]);
}
```

## Expected behavior

Keyprefix should not be ignored

wilhelmhb:

Using multiple translation functions is currently a pain point for me.

Yes #494 
Also provides a workaround for #973 
Should also fix #737 

### Checklist

- [x] documentation is changed or added (not relevant IMHO)


Bladieblah:
Merge pls!

karellm:
Can you please provide more context about this change? Keep in mind I maintain this project but haven't used it myself in years so I'm not aware of all the new use cases and syntax changes.

wilhelmhb:
@karellm 
As precised in #494, we sometimes want to fetch translations from different namespaces in one component.
For example, I have two common namespaces:
- `common`, with translations for "yes", "no", "continue", "cancel", etc
- `units` with translations regarding units: "per year", etc
Then I have (usually) one namespace/keyprefix for the current component.

Thus my components can start with something like:
```ts
const { t: tCommon } = useTranslation("common");
const { t: tUnits } = useTranslation("units");
const { t } = useTranslation("componentNamespace");
```
`i18next` supports this syntax without a hitch. However, if I have the following code in my component:

```ts
<span>{t("myField")}{value}{tUnits("perYear")}</span>
<button>{tCommon("confirm")}</button>
```

both `"perYear"` and `"confirm"` are either:
- added to the namespace `myComponent` if `tCommon` and `tUnit` are in the list of custom functions for extracting translations, which is pretty inconvenient.
OR
- ignored otherwise

I'm currently resorting to having my routing component (which do not need translations) with an *unused* translation function `t` defined + comments for each and every translation string in both the `common` and `units` namespaces to be extracted from.


Currently, the extractor keeps track of the **last defined** translation function, and adds detected translations in a file to the corresponding namespace and keyprefix.

With the changes, it keeps track of which translation function was defined which default arguments for namespace and keyprefix, and adds translations to the corresponding namespace and keyprefix accordingly.

I can provide you with a minimum working example if you so wish ;)

Pouet--:
I face this problem too and am interested with the merge :)

thomasmery:
Hi,

would love to this merge as well as it can be very difficult to understand why some keys are not extracted as they should

thanks

karellm:
Deployed as `9.3.0`

Pouet--:
@wilhelmhb : Thank you for the PR. However, I still have one problem, which is that every translation key in a file depends on the last keyPrefix used. Multiple Components with different useTranslation hooks, with different keyPrefix values, does not work :(

example :

```tsx
function Component1() {
  const { t } = useTranslation();
  ...
  { t('globalValue') }
}

function Component2() {
  const { t } = useTranslation(defaultNS, { keyPrefix: "user");
  ...
  <UserIcon />
  { t('account') }
}

function Component3() {
  const { t } = useTranslation(defaultNS, { keyPrefix: "admin");
  ...
  { t('somekey') }
}
```
 
Depending on the order components are defined, I got something like 
```
"user": {
  "account": "Compte",
  "globalValue": "Valeur globale",
  "admin": {
    "somekey": "Some key"
  }
}
```

EDIT : If I replace some of the `t` by `{t: tUser}` it definitely removes the key !

Pouet--:
@wilhelmhb Did you manage to get it work with your changes ?
Does anyone have a functional solution ?

wilhelmhb:
@Pouet-- you have to rename `t` functions to something else for the different key prefixes to be taken into account
\+ probably add the new function to the list of functions you extract translation keys from (see `functions: ['t']` in [the doc](https://github.com/i18next/i18next-parser?tab=readme-ov-file#javascript))
I haven't had the time to test it on my project, but tests are part of this PR, I encourage you to take a look at them, for they should be enough to adapt your code

Pouet--:
Thank you for your response. It does not make sense to me to rename my 't' functions as I am using one function per component, so there is no naming conflict on my side and I think it can lead to a more restricted programming experience. Thank you for your feedback anyway

rocky-luo-opus:
@Pouet-- 
Hello, I face the similar problem [1137](https://github.com/i18next/i18next-parser/issues/1137)
Do you have any methods to resolve this? Thanks!

Pouet--:
I unfortunately gave up because there is no one maintaining this repository :/
A method I tried was to rename all my "t" functions to avoid the library to only use the last one.

For example : 

```tsx
function Component1() {
  const { t: t1 } = useTranslation();
}

function Component2() {
  const { t: t2 } = useTranslation();
}
```

PR #1091, opened 2024-11-28
Add namespace information to duplicate warnings

janpaepke:

The warning message for duplicate keys is of limited usefulness, if it concerns a key that exists in multiple namespaces.
For example: the `title` key exists in 10 namespaces. When I get a warning that it's duplicate, I likely have accidentally used the wrong namespace somewhere, but I have no way of identifying where.

This PR Fixes this by adding the namespace to the log output.

No #000

### Checklist

- [x] tests are included and pass: `yarn test` (see [details here](https://github.com/i18next/i18next-parser/blob/master/docs/development.md#tests)) (no tests relevant)

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -186,10 +186,16 @@ export default class i18nTransform extends Transform {
           if (conflict === 'key') {
             this.warn(
               `Found translation key already mapped to a map or parent of ` +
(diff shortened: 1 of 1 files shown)


karellm:
Thanks!

PR #1093, opened 2024-12-09
Add support for i18next v24

### LINKED ISSUE #1090: Add support for i18next v24
## 🚀 Feature Proposal

v24 was released some days ago: https://github.com/i18next/i18next/releases/tag/v24.0.0

Migration guide: https://www.i18next.com/misc/migration-guide#v23.x.x-to-v24.0.0

The changes seem limited, so I hope it's not too hard to update. Maybe it might even be possible to support both versions in case the dropping of old browser support is an issue? It's not an issue for me at least.

## Motivation

I'd like to update to v24, but i18next-parser is blocking the update.

## Example

npm update :)

lazka:
v24 no longer supports compatibilityJSON=v3 so skip the tests for it when running with i18next v24.

Fixes #1090

See title

Yes #1090

### Checklist


package.json
```diff
@@ -36,7 +36,7 @@
     "esbuild": "^0.23.0",
     "fs-extra": "^11.2.0",
     "gulp-sort": "^2.0.0",
-    "i18next": "^23.5.1",
+    "i18next": "^23.5.1 || ^24.0.0",
```
ripecosta, line 39: Would it make sense to stop supporting v23 entirely instead?
If so, since that would be a breaking change, would it make sense to also move `i18next` to a peerDependency? It would help upstream projects with their dependency management.

lazka: > Would it make sense to stop supporting v23 entirely instead?

Your call. I seemed easy enough to support both, and it would make my life easier as I have various packages using i18next that depend on this, so I could update things gradually. But I'd be happy to change things if wanted.

> If so, since that would be a breaking change, would it make sense to also move `i18next` to a peerDependency? It would help upstream projects with their dependency management.

I don't quite understand how peerDependency could help as I've never worked with that. Note that i18next-parser is importing/using i18next directly and is not plugin as noted here: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependencies but maybe I'm missing something.

PR #1094, opened 2024-12-12
feat: HTML lexer supports vue bind attributes

PBK-B:

in Vue, there are v-bind-related attributes that can write simple syntax logic directly on the template syntax. This kind of code syntax exists in many traditional vue projects. I believe people will like that `t('translate')` written directly in v-bind can be recognized by i18next-parser.

vue sample code:
```vue
<template>
  <button :aria-label="t('b_a_l', 'button aria label')">
    button label
  </button>
  <button v-bind:aria-label="$t('button v-bind aria label')">
    button label form v-bind
  </button>
</template>
```

the solution I tried is to directly transform the existing html-lexer by adding `vueBindAttr: Boolean` option to support the acquisition of `:attribute="t('translate')"` and `v-bind:attribute="t('translate')"` syntax sugar in the html tag, and then parse the functions in it through javascript-lexer, and finally get the desired results. This scheme does not introduce additional dependencies but will make html-lexer dependent on javascript-lexer when the function is enabled.

usage example:
```js
export default {
……
    // see below for more details
    lexers: {
        vue: [
            {
                lexer: 'JavascriptLexer',
                functions: ['t', '$t'], // Array of functions to match
                namespaceFunctions: ['useTranslation', 'withTranslation'],
            },
            {
                lexer: 'HTMLLexer',
                functions: ['t', '$t'],
                vueBindAttr: true,
            },
        ],
        ……
    }
……
}
```

finally: @karellm TBR; At the same time, thank you for providing such a cool tool to save us a lot of time. I wish you a good day. If there is anything that can help promote this PR merger, welcome to comment.

#617 

### Checklist


karellm:
@PBK-B  Thank you for the PR and kind words. Would you be open to create a separate vue lexer that extends the html-lexer? I'm a bit reluctant to add framework specific code to the html-lexer. 

Also, I'm confused as to why we need to define both lexers in the config file. Is that necessary?

PBK-B:
@karellm I seem to have seen that vue-lexer was deleted just now (although I can understand that it is not dependent on vue-template-compiler)
It is not impossible to add a new vue-lexer, but I would like to ask for your thoughts on implementation, because the current changes are only based on html-lexer to add some small features, which itself still has to go through html-lexer first. Scan with javascript-lexer. If I create a new vue-lexer, is it acceptable for me to directly copy a copy of html-lexer?

secondly, configure both javascript-lexer and html-lexer for vue because the vue template syntax code file contains both javascript code and html-like code syntax.
a complete .vue code file looks like this:
```vue
<template>
  <button :aria-label="t('b_a_l', 'button aria label')">
    button label
  </button>
  <button v-bind:aria-label="$t('button v-bind aria label')">
    button label form v-bind
  </button>
  <p>{{t("tips", "this is an update prompt")}}</p>
  <p>{{msg}}</p>
  <p>{{elText}}</p>
</template>

<script lang="ts" setup>
import { computed, ref } from "vue"
import { t } from '@/i18n'

const msg = t("msg_text", "check for version updates")
const elText = computed(() => {
    return getVerisonUpdate() ? t('new_version', "new version") : t("not_new_version", "no updates available")
})
</script>
```

PBK-B:
In addition, I am worried that adding back the vue-lexer that was originally removed and implementing inconsistent functions will cause confusion when people upgrade the version.

PBK-B:
@karellm I moved the code out separately and created a `VueTemplateLexer` (to avoid differentiation from the old VueLexer), calling HTMLLexer and JavascriptLexer in extract and supporting VueTemplate `v-bind` syntax, If you have any good suggestions, please contact me

src/lexers/vue-template-lexer.js
```diff
@@ -0,0 +1,49 @@
+/*
+ * @Author: Bin
+ * @Date: 2025-01-04
+ * @FilePath: /i18next-parser/src/lexers/vue-template-lexer.js
+ */
+import HTMLLexer from './html-lexer.js'
+import JavascriptLexer from './javascript-lexer.js'
+import * as cheerio from 'cheerio'
+
+export default class VueTemplateLexer extends HTMLLexer {
```
geertklop, line 10: question: Why are you extending the HTMLLexer and then compose the jsLexer as well. For better maintainability, shouldn't it be extending from the BaseLexer and compose both the JSLexer and the HTMLLexer?

src/lexers/vue-template-lexer.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // support vue bind attribute
+    if (this.vueBindAttr) {
+      // traverse all tags to filter bind attribute
+      $('*').each((_, node) => {
+        const attributes = node.attributes.filter(
+          (item) => item.name.startsWith(':') || item.name.startsWith('v-bind:')
+        )
+        if (attributes.length > 0) {
+          // there are calculation attributes.
```
geertklop: Could you explain a bit more what the 'calculation' means?

geertklop (changes requested):
Great work @PBK-B

src/lexers/vue-template-lexer.js
PBK-B: Yes, by expanding BaseLexer, you can theoretically get better maintainability, but you will eventually find that it is actually a re-realization of HTMLLexer and JSLexer (excellent template syntax 🐶)

PBK-B: @geertklop Yes, by expanding BaseLexer, you can theoretically get better maintainability, but you will eventually find that it is actually a re-realization of HTMLLexer and JSLexer (excellent template syntax 🐶)

src/lexers/vue-template-lexer.js
PBK-B: @geertklop I'm very sorry, the misunderstanding caused by my colloquial translation should actually be calculation attributes = [Computed Properties](https://vuejs.org/guide/essentials/computed)

PBK-B:
@karellm The master branch has been rebased. If there is anything I can do to promote its merger, I am very willing to work for it (so much so that I have to distribute a package by myself in the project at present). Thank you for your work.

src/lexers/vue-template-lexer.js
geertklop: I still think that with the maintainability, I would expect to extend from BaseLexer. However, curious what regular maintainers like @karellm think.

PBK-B: @geertklop If you want, you can complete the follow-up work based on my branch

timothy-kumm:
@karellm Hey, we need this 🗡️

evanbattaglia:
I'm new to Vue and i18next, but I didn't see an obvious workaround for this feature not being implemented. The parsing was a major motivator in me using i18next, and I didn't want incomplete extraction, so it didn't make sense for my Vue project to use i18next without PBK-B's fork. I had to fork it to add node 24 to package.json, so it would be great to get this into master.

PBK-B:
Guys, because the project is abandoned and the official recommendation is to migrate to i18next-cli. So I will continue to work for this in the new project <https://github.com/PBK-B/i18next-cli-vue>. If it is helpful, please don't skimp on your star. Thank you very much (the PR will not be closed for the time being, so that Help the follow-up friends).

PR #1102, opened 2024-12-30
chore: Update lockfile

felix-xeel:

After some recent updates, the yarn lockfile needs to be updated.
`yarn install --frozen-lockfile` currently fails on `master`

No

### Checklist


karellm:
Should be fixed as of `9.2.0`

Issue #1103, opened 2025-01-03
Support Nested Property Namespace Functions in namespaceFunctions Lexer Option


gobzila:
## 🐛 Bug Report

Currently, the `namespaceFunctions` option in the lexers does not support recognizing wrapped functions defined as namespace properties, such as `i18n.useTranslate`. For example:

```
const { t } = i18n.useTranslate('common');
const message = t('welcome_message');
```

The parser fails to recognize `i18n.useTranslate` as a valid namespace function, and as a result, it does not extract the `common` namespace or the associated keys. The current implementation of `namespaceFunctions` only supports simple function names like `useTranslate`, which limits the parser's flexibility for projects that use wrapped functions defined within namespaces.

## Expected behavior

Extend the `namespaceFunctions` option in the lexers to support namespace property functions (e.g., `i18n.useTranslate`) in the same way that `functions` and `componentFunctions` were enhanced in [Issue #912](https://github.com/i18next/i18next-parser/issues/912) and [PR #913](https://github.com/i18next/i18next-parser/pull/913). This enhancement will allow the parser to extract namespaces and keys from nested property functions seamlessly.

## Why It's Important

This enhancement would improve the flexibility of the parser, making it compatible with projects that use namespaced or wrapped translation functions. It would also bring consistency across `functions`, `componentFunctions`, and `namespaceFunctions`, enabling developers to extract translations without requiring additional preprocessing or workarounds.

## Your Environment

- *runtime version*: node v20.11.1
- *i18next version*: 23.15.1
- *os*: Mac

karellm:
Should be fixed as of `9.2.0`

PR #1104, opened 2025-01-03
Support nested property namespace functions in namespaceFunctions

gobzila:

This PR adds support for recognizing namespace functions defined as nested properties (e.g., `i18n.useTranslate`) in the `namespaceFunctions` lexer option. This functionality brings consistency with functions and componentFunctions, which already support nested property functions.

Yes - [#1103](https://github.com/i18next/i18next-parser/issues/1103)

### Checklist

- [ ] documentation is changed or added - I believe the existing documentation is sufficient, but I'm happy to add more if needed.


karellm:
Thanks! Included in `9.2.0`

PR #1109, opened 2025-01-27
Update package.json: remove node restrictions

antonmedv:

To be able to use with node23.

No

### Checklist


karellm:
We've had issues in the past with this package not working with new releases of node. That's why the versions are limited. Until there is an active maintainer that takes over and will make sure non LTS versions work, I don't want to change this.

PR #1140, opened 2025-05-16
Add support for Node.js 24 and drop unused fields

LitoMore:

Node.js 24 is out, so we can consider adding Node.js 24 to the engines field.

The `npm` and `yarn` fields in `engines` can be removed since those versions are too old. We can drop them gracefully.

I also updated the GitHub Actions workflow file to test with Node.js 24.

No

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "i18next": "./bin/cli.js"
   },
   "engines": {
-    "node": "^18 || ^20 || ^22 || ^24"
+    "node": ">=18"
   },
   "scripts": {
     "coverage": "c8 --all --include='src/**/*[.js|.jsx|.ts|.tsx]' --reporter=lcov --reporter=text yarn test",
--- a/package.json
+++ b/package.json
@@ -11,9 +11,7 @@
     "i18next": "./bin/cli.js"
   },
   "engines": {
-    "node": "^18.0.0 || ^20.0.0 || ^22.0.0",
-    "npm": ">=6",
-    "yarn": ">=1"
+    "node": "^18 || ^20 || ^22 || ^24"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,9 +11,7 @@
     "i18next": "./bin/cli.js"
   },
   "engines": {
-    "node": "^18.0.0 || ^20.0.0 || ^22.0.0",
-    "npm": ">=6",
-    "yarn": ">=1"
+    "node": "^18 || ^20 || ^22 || ^24"
```
ripecosta: We should probably update it to ">=18" so people using uneven version don't have to use workarounds to install the library

LitoMore: You can still install the library with uneven versions. You will just receive a warning from npm like this:

ripecosta: That depends on the package manager. Some warn, others block installation. Yarn, for example, blocks installation.

LitoMore: I see. Then I think `>=18` will be better.

saitheexplorer:
Any updates on this? Trying to install another package that has this as a dependency and yarn is blocking the install because we're running node v24

DannyvanHolten:
Can we please merge this :)

sharknoon:
I am also running into the same `Unsupported engine` warning. All my other dependencies already support node.js 24 (which will become LTS in october 2025), ecxept the `i18next-parser`.

grafik

adrai:
fyi: https://github.com/i18next/i18next-parser/issues/1148

package.json
```diff
@@ -11,9 +11,7 @@
     "i18next": "./bin/cli.js"
   },
   "engines": {
-    "node": "^18.0.0 || ^20.0.0 || ^22.0.0",
-    "npm": ">=6",
-    "yarn": ">=1"
+    "node": ">=18"
```
recursive-franciscop, line 14: This is great since Node is very unlikely to break support for the APIs needed by this library, and this way it doesn't need continuous updating from the repo owner 🙏

PR #1146, opened 2025-08-06
fix(types): Mark defaultValue arguments required

andersk:

It seems these arguments are always provided to the `defaultValue` function.  (This doesn’t prevent the function from ignoring them: for example, `{ defaultValue: () => "hello" }` still satisfies this type.)

No

### Checklist


adrai:
fyi: https://github.com/i18next/i18next-parser/issues/1148

